PR #1, opened 2018-08-11
enforce prefix on car format to make future versioning easy

whyrusleeping:
Building cars this way, using tar, will always result in a specific header to the file. We can use this to distinguish 'car v0' from later versions of car.

### DIFF
--- a/car.go
+++ b/car.go
@@ -2,6 +2,7 @@ package car
 
 import (
 	"archive/tar"
+	"bufio"
 	"context"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


Stebalien:
Is there any way to inject a custom "this is a car" header so we don't try to interpret other TAR files as CARs?

Stebalien:
Ah. Nevermind, I see what's happening here. Is it too late to rename `root` to `car-root`?

Stebalien:
The other question is, should we *require* that root comes first? Given that this file *is* a valid TAR file, I'm worried that users will untar/retar and wonder why it doesn't work.

Stebalien:
Deprecated in favor of a custom format.

PR #13, opened 2019-09-19
add writecar method that avoids traversing a passed set of CID's

frrist:
### Description
This PR adds a method `WriteCarIgnoreSet` which accepts a set of cids that will not be traversed when writing a car file. 

### Status
If the addition of this method is well received tests will be added next.


frrist:
We found a work around that no longer requires this change cc @anorth.

PR #18, opened 2019-12-03
feat(car): add selector support

### LINKED ISSUE #14: Filtered DAG traversal (with IPLD selector?)
The go-filecoin project wanted to export a blockchain structure. The block headers have "State root" CIDs which refer to HAMT structures, which happen to be stored in the same store. We don't want to export the state trees though - i.e. we don't want to traverse those links. Similarly there is a "receipts" structure referenced from a block header that is not to be exported.

The CAR writer assumes that a complete DAG is to be exported, but we need flexibility. For expedience, we just did our own external traversal and write the CAR via the exported utility methods. A few different options would have been more helpful:
- a writer that consumes an iterator over blocks to write
- a writer that accepts a predicate against which each CID/block is to be tested before loading/writing
- a writer that accepts an IPLD selector describing the sub-DAG to write

That last option would be awesome (I know the selectors did not exist at the time this code was authored).

FYI @frrist @warpfork

hannahhoward:
# Goals

fix #14 

# Implementation

This adds the ability to write car files that contain only partial DAGs that start from their roots, using selectors.

```golang
var ctx context.Context
var bs car.ReadStore
var rootCid cid.Cid
var selector ipld.Node

sc := car.NewSelectiveCar(ctx, bs, []Dag{Dag{Root: rootCid, Selector: selector}})
```

It also adds a number of abilities that can be used while writing car files this way.

- You can estimate the size of the car file and get a unique list of CIDs that will go in the car file without actually writing it

```golang
scPrepared, err := sc.Prepare()
if err != nil {
  return err
}

fmt.Printf("Total size of car file when I write it will be: %d\n", scPrepared.Size())
fmt.Println("List of unique CIDs that will go into car file are:")
for _, c := range scaPrepared.Cids() {
  fmt.Println(c.String())
}

// fast write from prepared file
var w io.Writer
err := scPrepared.Dump(w)
```

- You can get metadata (via a hook) about blocks that are written to the car file when writing in a single step. (i.e. size and offset within CAR file)

```golang
var w io.Writer
// note: hook is actually var arg and is optional
err := scPrepared.Write(w, func(block car.Block) error {
   fmt.Printf("Next Block is Cid %s, at offset %d in file, and %d bytes long\n", block.BlockCID.String, block.Offset, block.Size) // car.Block also has block.Data for all the raw bytes for the block itself)
}
```


car.go
```diff
@@ -79,13 +83,22 @@ func ReadHeader(br *bufio.Reader) (*CarHeader, error) {
...
 	hb, err := cbor.DumpObject(h)
 	if err != nil {
 		return err
 	}
 
-	return util.LdWrite(cw.w, hb)
+	return util.LdWrite(w, hb)
+}
+
+func SizeHeader(h *CarHeader) (uint64, error) {
```
rvagg: maybe `HeaderSize()`? I know that `SizeHeader` fits in with the other `*Header()` names but "_something_ size" seems like it might make for an easier-to-grok API? otherwise this could be read as something like "a header containing this size"

selectivecar.go
```diff
@@ -0,0 +1,251 @@
...
+		return nil
+	}
+	size, err := sc.traverse(onCarHeader, onNewCarBlock)
+	if err != nil {
+		return SelectiveCarPrepared{}, err
+	}
+	return SelectiveCarPrepared{sc, size, header, cids}, nil
+}
+
+func (sc SelectiveCar) Write(w io.Writer, userOnNewCarBlocks ...OnNewCarBlockFunc) error {
```
rvagg: doc comment would be good here

rvagg: what's the purpose of being able to pass in multiple callbacks here?

rvagg: also, multiple callbacks should be tested too if it's not too hard

selectivecar.go
```diff
@@ -0,0 +1,251 @@
...
+type SelectiveCarPrepared struct {
+	SelectiveCar
+	size   uint64
+	header CarHeader
+	cids   []cid.Cid
+}
+
+// NewSelectiveCar creates a new SelectiveCar for the given car file based
+// a block store and set of root+selector pairs
+func NewSelectiveCar(ctx context.Context, store ReadStore, dags []Dag) SelectiveCar {
```
rvagg: @mikeal considering the next iteration of CAR, should we start making new APIs only take a single root rather than allowing for an array of roots? I can't recall exactly now but I think we were all agreeing that we should make it strictly one root and make the user work around that.

selectivecar.go
```diff
@@ -0,0 +1,251 @@
...
+			return nil, err
+		}
+		sct.offset += size
+	}
+	return bytes.NewReader(raw), nil
+}
+
+func (sct *selectiveCarTraverser) traverseBlocks() error {
+
+	nbc := dagpb.AddDagPBSupportToChooser(func(ipld.Link, ipld.LinkContext) ipld.NodeBuilder {
```
rvagg: is this something that should go in an `init`?

car_test.go
```diff
@@ -69,3 +73,79 @@ func TestRoundtrip(t *testing.T) {
...
+	blockCount := 0
+	err := sc.Write(buf, func(block Block) error {
+		blockCount++
+		return nil
+	})
+	require.Equal(t, blockCount, 5)
+	require.NoError(t, err)
+
+	// write car in two steps
+	scp, err := sc.Prepare()
```
rvagg: should we reset this to a new `NewSelectiveCar()` since this is testing an entirely separate workflow?

car_test.go
```diff
@@ -69,3 +73,79 @@ func TestRoundtrip(t *testing.T) {
...
+	nd3 := &dag.ProtoNode{}
+	nd3.AddNodeLink("second", nd2)
+	nd3.AddNodeLink("bear", c)
+
+	assertAddNodes(t, dserv, a, b, c, nd1, nd2, nd3)
+
+	ssb := builder.NewSelectorSpecBuilder(ipldfree.NodeBuilder())
+	selector := ssb.ExploreFields(func(efsb builder.ExploreFieldsSpecBuilder) {
+		efsb.Insert("Links",
+			ssb.ExploreIndex(1, ssb.ExploreRecursive(selector.RecursionLimitNone(), ssb.ExploreAll(ssb.ExploreRecursiveEdge()))))
```
rvagg: for future maintainers, could you add a small description of what this selector is supposed to do with the graph starting at `nd3`?

selectivecar.go
ribasushi: Note that go-ipfs will need to continue accepting multiple roots / no-roots / incomplete DAGs due to non-fil specific use-cases. I.e. keep the door open to "sloppy" or "strict=false" interfaces when making modifications.

rvagg: (a) let's enumerate those use-cases because it informs CARv2
(b) this is only for write interfaces, and afaik your exporter is producing CAR files with roots - can they be more than one root?

Let's get CARv2 discussions going before you ship go-ipfs with CAR import/export because I don't want us to lock in CAR to supporting more than we want it to support into the future.

ribasushi: For the foreseable future go-ipfs will *export* ( produce ) .car files with a single root, with the entire requisite DAG being present in the produced file. I.e. the only cars go-ipfs will be producing at lest in v0.1 are "fil-compliant .car's"

However the .car *import* functionality is nothing more than a generalization of `ipfs block put`, unblocking a plethora of use-cases currently relying on the inefficient one-at-a-time interface.

In other words: if go-car ( the library ) starts validating a .car file before "streaming it out" - there needs to be a way for the caller to disable these checks, because go-ipfs must be able to import:
- a .car with disjoint blocks and no roots
- a dag split over multiple .car files with the root specified only in one of the headers
- a mixture of the above with multiple roots specified in multiple headers
- roots specified in headers *without* the entire dag being contained  in *any* of the supplied .car files ( supplementary / update )

Let me know if this is not sufficient description and we could jump on a call / get a more proper doc going.

car_test.go
hannahhoward: that's reasonable.

car.go
hannahhoward: sure

selectivecar.go
hannahhoward: I would prefer not to try to incorporate things that anticipate the next version of CAR until it settles, and then adjust then.

hannahhoward:
there is definitely a use case in having both modes and it's actively used in filecoin. The use case is:
- I want to generate commP
- I want to do so using os.Pipe so I never put the whole CAR on disk
- I need to know the pieceSize ahead of time

regarding API surface area:
- I'm happy to fold this back in when CAR spec v2 lands
- that has no clear date
- Folks in filecoin land are actively using this code and need to use this code.

hannahhoward:
arguably, we could just break it into it's own repo

selectivecar.go
rvagg: yeah ok, so I chatted with @mikeal about this and he still sees the multi-root use-case sticking around and that'll probably end up in CARv2, so we can move this discussion to ipld/specs when I get around (soon) to doing some initial CARv2 work.

rvagg:
> arguably, we could just break it into it's own repo

it seems that a "is deterministic" flag and "[root + selector]" are desired features for CARv2, so this all makes sense in here and maybe should be encouraged as the primary interface in a future version.

rvagg:
Streaming CommP from a blockstore is a use-case I definitely understand!

Let's just address those minor tweaks and get this landed.

PR #20, opened 2020-02-21
Use correct walk func in enumGetLinks

magik6k:


magik6k:
Depends on https://github.com/ipfs/go-car/pull/20

anorth:
See #14

PR #21, opened 2020-02-25
Add ignore paaram to WriteCar for cids we don't want to write

ZenGround0:
Filecoin requirements put us in the position of needing to ignore certain cids (huge sector data graphs) while writing dags to carfiles.  This adds a parameter for ignoring cids in a graph during the write traversal.


ZenGround0:
Closing in favor of #20

PR #24, opened 2020-03-04
Support filestore when reading CAR file

hannahhoward:
# Goals

If the reader given to LoadCar is a file (i.e. files.FileInfo) and the Store is a filestore, support writing to the filemanager rather than the blockstore, so that no extra data is imported to the local blockstore

# Implementation

- modify LdRead to return total length read
- track offset in file
- write posinfo if neccesary when reading CAR file
- write test to verify we can rewrite car file without ever writing to into intermediate blockstore


car.go
```diff
@@ -159,7 +171,25 @@ func (cr *CarReader) Next() (blocks.Block, error) {
...
 
-	return blocks.NewBlockWithCid(data, c)
+	offset := cr.offset + l - uint64(len(data))
+	cr.offset += l
+
+	blk, err := blocks.NewBlockWithCid(data, c)
+	if cr.fullPath == "" {
+		return blk, err
+	}
+	nd, err := format.Decode(blk)
```
rvagg: From what I can tell this is going to force a full decode of the data using the appropriate codec, which is a new piece of functionality for go-car and possibly a significant new cost. I don't know why `Filestore` needs a full `Node` since it's only touching `RawData()` and `Cid()`, the elements of `Block`, so we're being forced here to decode a full Node just because the Filestore API wants the whole-hog?

This seems like a significant cost overhead to me, just for the use-case of being able to write to a Filestore which may not be a majority use-case. Am I right? Got any thoughts on how we might make this more efficient for the common use-case or even wipe away this decode entirely?

rvagg:
There's a `ReadHeader` in car/main.go that needs to be adjusted to accept the 3 return values.

Also worth nothing that this would be a breaking change since we expose `ReadHeader` and `LdRead` whose signatures would change. I'm not yet sure how to think about breaking changes in this ecosystem, but they are annoying regardless of ecosystem.

The Filestore API is interesting, I like that it can do a "binary data is over there in that file at this location", but setting it up seems ridiculously expensive--all it really needs to be is `Put(Cid, PosInfo)` but you're now doing a full decode which is very unlikely to be consumed anywhere. Would like to hear your thoughts on whether you agree that this is an extra cost (maybe I'm missing something) and whether we can deal with that. Maybe we have to push it back up to Filestore to get it to accept simplified input.

aschmahmann:
IIUC this PR is going to give us problems in go-ipfs land. Go-IPFS is using `go-filestore` pre version 1.0.0 because we are not yet using multihashes in the blockstore, This means that we will not be able to import the latest version of `go-car` into go-ipfs without everything breaking.

cc @Stebalien @hsanjuan @jacobheun

Stebalien:
Honestly, that's probably fine. If it's a problem, we can patch it.

However, the simple solution here is to simply have this depend on a pre-1.0 version of the filestore. That should work for everyone.

mvdan:
This PR also contains support for null padding when reading files. I'd say keep that out for now, because I want to add that as an opt-in feature; see https://github.com/ipld/go-car/issues/171. It should be opt-in, because otherwise it's a forced change of behavior that goes against the spec, and a potential breaking change.

Also, in terms of API, I see that this changes the signature of `car.ReadHeader` and `util.ReadNode`. Did you consider ways to add the feature without making breaking changes to the API? I know we're in a v0 and we're technically allowed to break users, but it's still best to avoid doing so.

rvagg:
As per https://github.com/ipld/go-car/issues/254#issuecomment-1122215966, I believe this branch is no longer in use by lotus, directly or indirectly. Closing this PR now but the branch can remain.

rvagg:
since it's been included in so many historical Lotus (and other) releases,`feat/filestore` has been added as a protected branch, so future cleanups shouldn't impact it

PR #25, opened 2020-03-04
don't use private types in public functions

Stebalien:
Also, remove some dead code.


car.go
```diff
@@ -105,12 +105,12 @@ func (cw *carWriter) writeNode(ctx context.Context, nd format.Node) error {
...
 }
 
-type carReader struct {
+type CarReader struct {
 	br     *bufio.Reader
 	Header *CarHeader
 }
 
-func NewCarReader(r io.Reader) (*carReader, error) {
+func NewCarReader(r io.Reader) (*CarReader, error) {
```
Stebalien: Otherwise, the user can't store this.

PR #28, opened 2020-03-27
feat: only require the NodeGetter interface instead of the DAG interface

Stebalien:
This makes it possible to pass in a dagservice "session" for better performance when fetching from bitswap.


ribasushi:
Testing with changes in go-ipfs, will know results in a bit...

Stebalien:
Probably, yes.

PR #31, opened 2020-04-29
Update go-ipld-prime to the era of NodeAssembler.

warpfork:
Bumps dependencies and updates a few interfaces and package names that have shifted lately.

`go mod` did a lot of other things that make very little sense to me -- the two commands I ran were `go get github.com/ipld/go-ipld-prime@master` and `go get github.com/ipld/go-ipld-prime-proto@master`, and how that caused `go-ipld-format` to show up in the update effects (it's a test dependency only of a transitive lib, afaict?), or why lots of other things appear in the the 'go.sum' file that are even more distantly (un)related and also get many versions recorded which we haven't actually handled... I... I don't know.  Someone may want to sanity check that; it's clear I simply don't understand why `go mod` is trying to do, and probably disagree with it in any case.  But tests pass; so, I think the update overall is passable.


warpfork:
Fascinatingly, that:
- networked, again (which I would hold is fundamentally inappropriate for the requested operation)
- removed some things
- added even more things???? (if these hashes weren't encountered or important *before*, how did they become encountered, much less become *important*, during "tidy"????)

I've committed the results.  However, these behaviors strike me as not so much minor bugs as they are boggling infirmities of core concepts -- seeing a system that appears to use hashes, but also does so in such inconsistent ways that one begins to wonder if those hashes are being *checked* or actually *mean anything* is deeply disquieting and something I have an especially strong allergy to.  I formally give up trying to reason about go mod.  It does not spark joy.

PR #32, opened 2020-05-26
Handle reading null-padded files

magik6k:


rvagg:
Wait, what? Can you explain what this is doing and what it's here for? It looks like this is a spec adjustment. It might be good to hold on PRs like this for a decent amount of time for review since we now have a lot of moving pieces depending on CAR implementations being stable and this isn't the only CAR implementation in the wild and being used.

rvagg:
> merged commit 3b87370 into `ipld:feat/filestore`

OK, sorry, this wasn't into master, not a big deal then. But could we have a clearer rationale and discussion about the implications so we can move on with the other PR now that it's being further complicated? My guess is this is to support reading from Filecoin sectors where the CAR contents get zero padded at the end to get to the right size for storage, but doesn't Filecoin store length information about chunks that would allow it to avoid pushing this concern down into CAR decoding? It seems like something that belongs at a layer or two above here?

PR #36, opened 2020-11-05
Improve EOF handling

rvagg:
This came up from go-ipfs folks trying to import a CAR file that was truncated but they had no indication that it was. The final block had an `0xa0` after it, which is an improper varint, so `ReadUvarint()` was returning an `io.EOF` which also happens if there's _no_ next block at all and is therefore the proper-EOF signal. This change lifts the proper-EOF case out into a `Peek(1)` so every other EOF from there on can be assumed to be unexpected.

### DIFF
--- a/car/main.go
+++ b/car/main.go
@@ -26,7 +26,7 @@ var headerCmd = cli.Command{
 		}
 		defer fi.Close()
 
-		ch, err := car.ReadHeader(bufio.NewReader(fi))
+		ch, _, err := car.ReadHeader(bufio.NewReader(fi))
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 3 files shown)


mvdan (approved):
Looks great :) The CAR chunk of bytes would be a bit shorter if you used base64 by the way, via base64.StdEncoding.DecodeString.

rvagg:
thanks @mvdan, on the topic of bytes as strings: I prefer hex for for this kind of thing simply for its portability, this string for instance is straight out of JavaScript. Base64 tends to involve more dependencies and hoop-jumping in various languages, while hex is usually lowest-common-denominator

PR #37, opened 2020-11-17
Allow user defined block hooks when using two step write for selective cars

hannahhoward:
# Goals

This is a simple but pretty obscure feature. When building cars with Selectors, we have an option for a two step write -- one step performs the traversal, while another writes the traversed cids to a file. This is important among other things so that we can get a total file size ahead of time. We also have an option in the 1-step write to call a user defined hook as we write blocks. Now, we allow the two step write to use these user-defined hooks. However, they are not called till you actually dump the file. For a variety of reasons, we need this for Filecoin, but it seems to me to be a non-controversial enough change as to allow us to merge to master?

# Implementation

- hold the user funcs in the intermediate structure
- track offset/size and call during io.Write

### DIFF
--- a/car_test.go
+++ b/car_test.go
@@ -117,7 +117,9 @@ func TestRoundtripSelective(t *testing.T) {
 	// write car in one step
 	buf := new(bytes.Buffer)
 	blockCount := 0
+	var oneStepBlocks []Block
 	err := sc.Write(buf, func(block Block) error {
+		oneStepBlocks = append(oneStepBlocks, block)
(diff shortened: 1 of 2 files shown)


rvagg (approved):
sgtm

there's a loss of coverage here though because the `Write()` and `Prepare()` _without hook_ case isn't being covered anymore

hannahhoward:
I merged cause there's not a seperate code path for no-hooks -- it's just simpler

PR #46, opened 2021-02-11
add license file so it can be found by go-licenses

bollwyvl:
Thanks for this! I'm trying to package go-ipfs for [conda-forge](
https://github.com/conda-forge/staged-recipes/pull/13904), and this is one of the few modules that doesn't have a discoverable license file.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) Whyrusleeping
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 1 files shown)


rvagg:
They should actually be dual licensed (along with the other packages you're adding this to), so a LICENSE-MIT and LICENSE-APACHE file, like https://github.com/ipld/js-car. Is that going to be a problem for go-licenses?

bollwyvl:
Yes, i think all the magic license tools know how to find `LICENSE-*` just as well, and will bring in both... I saw that root IPFS was doing the dual license thing, but didn't see any reference to that on this repo. With Apache, a project really does have to have the license in the repo and distribute it with source and binaries to comply with the terms.

rvagg:
Yeah, it's a Protocol Labs policy to dual license its original projects, but it's regularly overlooked with some folks considering it _implied_ since it's documented in other places. That's probably not a great idea and getting them in explicitly would be good.

Stebalien:
Replaced with https://github.com/ipld/go-car/pull/57

Issue #53, opened 2021-03-08
disable Travis


marten-seemann:
This repository is using the unified Go testing / checking workflow. We should disable Travis (and delete `.travis.yml`).
I don't have permission to edit CI settings. @Stebalien, do you?

mvdan:
As far as I can tell Travis isn't running anymore, or at least isn't showing up in the CI reports on master commits. The travis.yml file was deleted.

marten-seemann:
Thanks for the ping. You're right, this issue was resolved at some point.

PR #54, opened 2021-03-08
replace go-ipld-prime-proto with go-codec-dagpb

mvdan:
(see commit message)


willscott (approved):

selectivecar.go
```diff
@@ -238,21 +241,22 @@ func (sct *selectiveCarTraverser) loader(lnk ipld.Link, ctx ipld.LinkContext) (i
...
 func (sct *selectiveCarTraverser) traverseBlocks() error {
-
-	nsc := dagpb.AddDagPBSupportToChooser(func(ipld.Link, ipld.LinkContext) (ipld.NodePrototype, error) {
+	nsc := func(lnk ipld.Link, lctx ipld.LinkContext) (ipld.NodePrototype, error) {
+		// We can decode all nodes into basicnode's Any, except for
+		// dagpb nodes, which must explicitly use the PBNode prototype.
+		if lnk, ok := lnk.(cidlink.Link); ok && lnk.Cid.Prefix().Codec == 0x70 {
+			return dagpb.Type.PBNode, nil
+		}
 		return basicnode.Prototype.Any, nil
```
hannahhoward, line 250: As I read this, a raw node will get built with an Any node, rather than a Bytes node? Will that matter?

hannahhoward (approved):
Curious about Raw nodes getting built with ProtoType.Any and whether that's an inefficiency, but otherwise LGTM.

selectivecar.go
mvdan: As far as I could tell, your RawNode was the same as a basicnode.Bytes at the codegen level, so it seemed unnecessary to expose an identic type in go-codec-dagpb again.

I'm pretty sure that the use of Any shouldn't incur an inefficiency, as it barely does much more than the Bytes prototype: https://github.com/ipld/go-ipld-prime/blob/6de9f5cbb6bd0d00246dac8cae896e8cb175029d/node/basic/any.go#L135-L142

But if the measurements prove me wrong, it wouldn't be hard to re-add a case here to return basicnode.Prototype.Bytes :)

mvdan:
The Travis failures can be ignored as per https://github.com/ipld/go-car/issues/53. Merging, as this shouldn't be a breaking change to anyone downstream.

PR #58, opened 2021-03-25
chore: remove LICENSE

Stebalien:
It looks like this may be confusing pkg.go.dev's license detection.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,0 @@
-This project is dual-licensed under Apache 2.0 and MIT terms.
-
-MIT: https://www.opensource.org/licenses/mit
-Apache-2.0: https://www.apache.org/licenses/license-2.0


Stebalien:
Note: this leaves LICENSE-MIT and LICENSE-APACHE

ribasushi:
@Stebalien just a random note: when you leave two license files like this, github ends up randomly selecting one of them for the built-in license-link stuff.
I settled on the following after latest changes to pkg.go.dev: single file, has intent, has both licenses, keeps everything happy: https://github.com/ipfs/ipfs-repository-template/blob/master/LICENSE.md

image

Stebalien:
Ah, does that work with https://pkg.go.dev/github.com/ipld/go-car? I removed this because go is displaying a scary "UNKNOWN".

But I assume the SPDX thing may fix it. I'll try that.

Stebalien:
Here: https://github.com/ipld/go-car/pull/59

PR #67, opened 2021-04-01
fix staticcheck errors

marten-seemann:


car.go
```diff
@@ -183,17 +183,16 @@ func loadCarFast(s batchStore, cr *CarReader) (*CarHeader, error) {
 	var buf []blocks.Block
 	for {
 		blk, err := cr.Next()
-		switch err {
```
marten-seemann, line 186: Not sure if there was a reason to use a `switch` here. Performance maybe, since this is a function that's supposedly "fast"?
If I remember correctly though, `switch` is not faster than if / else, unless you have **a lot** of cases.

mvdan: I honestly don't think it matters here, either.

For those reading this later, the warning is:

> default case should be first or last in switch statement (ST1015)

It's entirely reasonable. I also think the if/else method is simpler.

mvdan (approved):

PR #75, opened 2021-06-03
fix: ReadHeader return value mismatch

rvagg:
./car/ tends to get overlooked cause it has no tests in it, but the new Actions errors on problems in there now thankfully, see #74


car/go.mod
```diff
@@ -3,8 +3,8 @@ module github.com/ipld/go-car/car
 go 1.13
 
 require (
-	github.com/ipld/go-car v0.1.0
-	github.com/urfave/cli v1.22.4
+	github.com/ipld/go-car v0.3.0
```
willscott, line 6: bump go requirement to 1.15 while we're at it?

rvagg: sure, I actually removed go.sum and go.mod entirely (cause it seems like a reasonable way to "clean" things up) and recreated so manually put 1.13 back in there

rvagg:
pulled in #74 to resolve codecov errors, also fixed a few lint/vet/deprecation errors it throws up

marten-seemann (approved):

mvdan (approved):

car/go.mod
```diff
@@ -1,10 +1,10 @@
 module github.com/ipld/go-car/car
 
-go 1.13
+go 1.15
```
masih, line 3: I am curious why not upgrade to `1.16` while we are at it

masih (approved):

car/go.mod
mvdan: You can think of this version as the absolute minimum that is required to build the module. We generally support the two latest major Go versions, being 1.15.x and 1.16.x right now, so 1.15 is the highest we can set.

masih: I see; that makes sense. Thanks

PR #76, opened 2021-06-03
Header error tests

rvagg:
Also pulls in #74 and #75

Ref: ipld/js-car#28

> Also included in this is an explicit check for `0aa16776657273696f6e02` (`<length>{version:2}`) as per https://github.com/ipld/specs/pull/248#issuecomment-833141588 so it can be used as a CARv2 escape - if we end up using this string as a modifier for CARv2 we can at least get proper errors out of this if it encounters a newer format.

the primary difference with js-car is that go-car requires a non-empty roots array whereas js-car is fine with empty roots array, hence the test fixture differences

### DIFF
--- a/car_test.go
+++ b/car_test.go
@@ -237,6 +237,58 @@ func TestEOFHandling(t *testing.T) {
 }
 
 func TestBadHeaders(t *testing.T) {
+	testCases := []struct {
+		name   string
+		hex    string
+		errStr string // either the whole error string
+		errPfx string // or just the prefix
+	}{
+		{
+			"{version:2}",
+			"0aa16776657273696f6e02",
+			"invalid car version: 2",
+			"",
+		},
+		{
+			// an unfortunate error because we don't use a pointer
+			"{roots:[baeaaaa3bmjrq]}",
+			"13a165726f6f747381d82a480001000003616263",
+			"invalid car version: 0",
+			"",
+		}, {
+			"{version:\"1\",roots:[baeaaaa3bmjrq]}",
+			"1da265726f6f747381d82a4800010000036162636776657273696f6e6131",
+			"", "invalid header: ",
+		}, {
+			"{version:1}",
+			"0aa16776657273696f6e01",
+			"empty car, no roots",
+			"",
+		}, {
+			"{version:1,roots:{cid:baeaaaa3bmjrq}}",
+			"20a265726f6f7473a163636964d82a4800010000036162636776657273696f6e01",
+			"",
+			"invalid header: ",
+		}, {
+			"{version:1,roots:[baeaaaa3bmjrq],blip:true}",
+			"22a364626c6970f565726f6f747381d82a4800010000036162636776657273696f6e01",
+			"",
+			"invalid header: ",
+		}, {
+			"[1,[]]",
+			"03820180",
+			"",
+			"invalid header: ",
+		}, {
+			// this is an unfortunate error, it'd be nice to catch it better but it's
+			// very unlikely we'd ever see this in practice
+			"null",
+			"01f6",
+			"",
+			"invalid car version: 0",
+		},
+	}
+
 	makeCar := func(t *testing.T, byts string) error {
 		fixture, err := hex.DecodeString(byts)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


mvdan:
Hm, why not merge #75 first? otherwise the review gets tricky :)

car_test.go
```diff
@@ -234,3 +235,80 @@ func TestEOFHandling(t *testing.T) {
...
+
+	// this is an unfortunate error, it'd be nice to catch it better but it's
+	// very unlikely we'd ever see this in practice
+	t.Run("null", func(t *testing.T) {
+		err := makeCar(t, "01f6")
+		if !strings.HasPrefix(err.Error(), "invalid car version: 0") {
+			t.Fatalf("bad error: %v", err)
+		}
+	})
+}
```
masih, line 314: Thinking out loud, would it be easier to define an anonymous struct of test cases and `t.Run` over them in a `for` loop? That should make the tests less verbose and easier to expand the test cases.

.github/workflows/automerge.yml
```diff
@@ -5,11 +5,35 @@
...
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+    - name: Check if we should automerge
+      id: should-automerge
+      run: |
+        for commit in $(git rev-list --first-parent origin/${{ github.event.pull_request.base.ref }}..${{ github.event.pull_request.head.sha }}); do
+          committer=$(git show --format=$'%ce' -s $commit)
+          echo "Committer: $committer"
+          if [[ "$committer" != "web3-bot@users.noreply.github.com" ]]; then
```
masih, line 26: I _think_ emails can be spoofed in commits. I'm curious do we need to worry about it?

.github/workflows/automerge.yml
```diff
@@ -5,11 +5,35 @@
...
+            echo "Commit $commit wasn't committed by web3-bot, but by $committer."
```
masih, line 27: I was wondering; do we follow any specific bash style guide like [this](https://google.github.io/styleguide/shellguide.html)?

masih (commented):
I'm not entirely sure on the order of PRs that are open. If this pulls other PRs in, would it make sense to close the other ones?

Left a bunch of questions.  
LGTM 🍻

.github/workflows/automerge.yml
mvdan: Not that I'm aware. This is all linside YAML, so trying to follow a strict format is going to be tricky...

.github/workflows/automerge.yml
mvdan: FYI this is just copied automatically from https://github.com/protocol/.github. It's not a security measure, but rather just a sanity check to catch some human mistakes. @marten-seemann can perhaps expand on this, but you should raise any issues on that repo :)

.github/workflows/automerge.yml
masih: I agree; bash in yaml is, well, bash in yaml 😊

.github/workflows/automerge.yml
masih: Gotcha. I am curious is there a reason for combining PRs into one?

car_test.go
rvagg: It would be easier if I wasn't a noob and didn't have to google how to do this each time it gets suggested! Have done it in the latest commit, ptal.

.github/workflows/automerge.yml
rvagg: to get CI to pass, that's all merged now so this should be just the new bits

.github/workflows/automerge.yml
rvagg: better take that up over in https://github.com/protocol/.github I think

rvagg:
okie dokie, this is much simpler to review now, just the relevant changes and some .gitignore action

car_test.go
```diff
@@ -234,3 +235,88 @@ func TestEOFHandling(t *testing.T) {
...
+	t.Run("Sanity check {version:1,roots:[baeaaaa3bmjrq]}", func(t *testing.T) {
+		err := makeCar(t, "1ca265726f6f747381d82a4800010000036162636776657273696f6e01")
+		if err != nil {
+			t.Fatal(err)
+		}
+	})
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			err := makeCar(t, tc.hex)
```
masih, line 310: Should we check if `err` is not `nil`? Otherwise, I _think_ the code could  panic because of `nil` pointer.

rvagg: 👍 good call, have added an `err == nil` check because we always expect an error here and if there isn't one then it's an ... error

mvdan: FYI, another way to do this would be `fmt.Sprint(err)` instead of `err.Error`, then that will give you `<nil>` for nil errors

mvdan (approved):

.github/workflows/automerge.yml
marten-seemann: @masih See https://github.com/protocol/.github/pull/85#discussion_r625235315.

PR #78, opened 2021-06-09
Implement a draft of CAR v2 writer

masih:
Aims to pin down the exact format of a CAR v2 based on the discussions with folks and reading the draft spec documentation. 

- [x] Define constants a structs to capture the primitives in CAR v2.
- [x] Generate index using [carbs](https://github.com/willscott/carbs).
- [x] Implement a CAR v2 writer for a given slice of root CIDs, using the CAR v1 node traversal libraries.
- [ ] Add writer tests.

Relates to https://github.com/ipld/specs/pull/248#issuecomment-833141588


v2/car.go
```diff
@@ -0,0 +1,70 @@
...
+import (
+	"encoding/binary"
+	"io"
+)
+
+const HeaderBytesLen uint64 = 32
+
+var (
+	// The fixed prefix of a CAR v2, signalling the version number to previois versoions for graceful fail over.
+	PrefixBytes = []byte{
```
mvdan, line 12: probably best to not export these vars

v2/car.go
```diff
@@ -0,0 +1,70 @@
+package car
```
mvdan, line 1: is the plan to do a v2 dir inside the Go v1 module, or to do an entirely separate Go v2 module? assuming we want the latter, I think it would be best to target PRs to e.g. a wip-v2 branch, so we can gradually review+merge and track progress that way without messing with master or having an ever-growing PR.

v2/car.go
```diff
@@ -0,0 +1,70 @@
...
+	// Header represents the CAR v2 header/pragma.
+	Header struct {
+		// 128-bit characteristics of this CAR v2 file, such as order, deduplication, etc. Reserved for future use.
+		Characteristics Characteristics
+		// The length of CAR v1 bytes.
+		CarV1Len uint64
+		// The offset from the beginning of the file at which the CAR v2 index begins.
+		IndexOffset uint64
+	}
+	Characteristics []byte
```
mvdan, line 35: I seem to recall this is 128 bits, so I think using a fixed-size type would be more appropriate. you have multiple options:

```
[16]byte
[2]uint64
struct{hi, lo uint64}
```

that last one is what I see most commonly, e.g. it's often used for ipv6 which is also 128 bits. dealing with two underlying integers is easier than dealing with 16, and using fields is a tiny bit nicer than arrays.

v2/go.mod
```diff
@@ -0,0 +1,27 @@
...
+	github.com/stretchr/testify v1.7.0
+	github.com/whyrusleeping/cbor-gen v0.0.0-20210422071115-ad5b82622e0f // indirect
+	github.com/willscott/carbs v0.0.5-0.20210604075022-1036c89e1fd7
+	go.uber.org/multierr v1.7.0 // indirect
+	go.uber.org/zap v1.17.0 // indirect
+	golang.org/x/crypto v0.0.0-20210513164829-c07d793c2f9a // indirect
+	golang.org/x/sys v0.0.0-20210608053332-aa57babbf139 // indirect
+)
+
+replace github.com/ipld/go-car => ../
```
mvdan, line 27: have you given much thought to whether Go's v2 module should depend on the v1 module, or if it should stand on its own ground?

Having the dependency has the upside of less code duplication. the v1 module works, and if the APIs are going to be largely compatible, it's nice to just do it that way.

There are two main downsides I see to having v2 depend on v1: we'll now have to maintain two modules instead of one (e.g. coordinating v1 and then v2 releases), and if v2 needs to directly use any internal code from v1, you won't be able to given they are in different packages.

v2/writer.go
```diff
@@ -0,0 +1,86 @@
+package car
+
+import (
+	"bytes"
+	"context"
+	"github.com/ipfs/go-cid"
+	format "github.com/ipfs/go-ipld-format"
+	car_v1 "github.com/ipld/go-car"
+	"github.com/willscott/carbs"
```
mvdan, line 9: I imagine we eventually want to just move the carbs code into this git repo?

v2/car.go
```diff
@@ -0,0 +1,70 @@
+package car
+
+import (
+	"encoding/binary"
+	"io"
+)
+
+const HeaderBytesLen uint64 = 32
```
mvdan, line 8: no need for this to be a typed constant :) and perhaps no need to export it either, as it's only useful to someone encoding/decoding CAR themselves

v2/car.go
```diff
@@ -0,0 +1,70 @@
...
+	EmptyCharacteristics = Characteristics(make([]byte, 16))
+)
+
+type (
+	// Header represents the CAR v2 header/pragma.
+	Header struct {
+		// 128-bit characteristics of this CAR v2 file, such as order, deduplication, etc. Reserved for future use.
+		Characteristics Characteristics
+		// The length of CAR v1 bytes.
+		CarV1Len uint64
```
mvdan, line 31: nit: we might want to swap "len" with "size" in all these, because the latter is much more clear in terms of "number of bytes" for most people.

"length" can be a bit more ambiguous: I could interpret it as the string length (number of unicode runes), as the length of the roots list, as the length of the list of blocks...

v2/car.go
```diff
@@ -0,0 +1,70 @@
...
+// header for convenient chained calls.
+// The index offset is calculated as the sum of PrefixBytesLen, HeaderBytesLen,
+// Header#CarV1Len, and the given padding.
+func (h *Header) WithPadding(padding uint64) *Header {
+	h.IndexOffset = PrefixBytesLen + HeaderBytesLen + h.CarV1Len + padding
+	return h
+}
+
+// Marshal serializes this header to bytes and writes the bytes using the given io.Writer.
+func (h *Header) Marshal(w io.Writer) (err error) {
```
mvdan, line 56: nit: most Marshal/Encode APIs in Go take the object you're encoding, so it's like "Marshal X into writer W". Here, it almost seems like you're marshaling a writer.

Perhaps a more unambiguous name would be MarshalTo. A more standard version of that is https://golang.org/pkg/io/#WriterTo. It seems appropriate to me, because there's just one way to represent a Header in the form of []byte, so it's not like WriteTo could be an ambiguous method.

v2/writer.go
```diff
@@ -0,0 +1,86 @@
...
+
+// Write writes the given root CIDs according to CAR v2 specification, traversing the DAG using the
+// Writer#Walk function.
+func (w *Writer) Write(ctx context.Context, roots []cid.Cid) (err error) {
+	err = w.writePrefix()
+	if err != nil {
+		return
+	}
+
+	buf, err := w.encodeCarV1(ctx, roots)
```
mvdan, line 44: I think a bit of docs as to why we're buffering the entirety of the inner carv1 would be nice. it seems like the only reason is building the index.

v2/car.go
mvdan: another reason to prefer uint64 instead of uint8 as the "element" type underneath: you can use math/bits funcs that work on more bits at a time.

v2/car.go
masih: That makes sense; I'll reopen the PR agains a v2 branch. 

Re v2 module I share the same thoughts. This PR indeed defines a new v2 module

Thanks @mvdan

v2/go.mod
masih: If the aim is to minimise barrier for adoption, then I _think_ dependency is the path of least resistance. Open to what you folks think works best.

v2/writer.go
masih: yes definitely. I'll get things moved in the coming PRs.

v2/writer.go
masih: Well spotted. will do. 

For reference, like you said we need to read the entire CAR v1 to then generate the index. Another less important reason is to set the size in CAR v2 `Header`; with a `Write` API that takes root CIDs we need to generate the CAR v1 to know its size. As far as I can see CAR v1 APIs don't allow the size of a CAR to be calculated without encoding it.

v2/go.mod
mvdan: If that were the case, then I would assume we would not be doing a v2 Go module at all - we would just add new APIs or options to the v1 Go module.

From what Will said, I understood that we'd break away from the v1 module and do a v2 module, while keeping APIs that work with the old IPLD libraries to make upgrading a bit easier. upgrading then would, ideally, just mean adding v2 to the go-car import paths and doing small local tweaks.

It seems to me like what you're doing here is a middle ground, which kinda gets us the disadvantages of both approaches at the same time :) people on the v1 module will still have to edit their code to add support for the v2 module, and on our side we'll add work for ourselves by maintaining two modules instead of one.

mvdan: After a brief chat on Slack, I realised I wasn't thinking about existing carv1 users. They should continue to work with zero changes, ideally, which means we'd have to continue maintaining the v1 module anyway. So I'm leaning towards my comments above being wrong.

masih: Thank you for explaining; that makes sense.

The approach in this PR assumes that "small local tweaks" is not acceptable on the v1 users' codebase. The v2 would then return types from the v1 library, as well as providing a newer IPLD flavoured library. 

If we can get away with enforcing small local tweaks to users whom want to use v2 then great, I think what you are suggesting with a totally separate v2 module is much cleaner :)

@willscott what are your thoughts?

v2/writer.go
mvdan: If you just needed the size, you could still figure that out without buffering the entire thing thanks to the io.Writer interface. You could layer an io.Writer on top of `w` when encoding the carv1, which would simply forward the `Write` calls while also keeping an incremental count of the `n int` returned by each Write call, e.g. in a uint64. This is how you typically count the total or "current progress" amount of bytes that go through an io.Reader or io.Writer withotu buffering the whole thing, anyway.

v2/car.go
willscott: this seems like a reasonable direction

v2/car.go
willscott: this becomes the magic bytes for filetype detection, so there's a plausible reason another package might want to reference them.

v2/writer.go
willscott: but the size needs to be written before all the car bytes.

For efficiency, we should allow writing to a `WriterAt` where we can seek back and fill in the header after the carv1 gets streamed.

masih: aha great suggestion thank you

v2/car.go
```diff
@@ -0,0 +1,70 @@
...
+
+import (
+	"encoding/binary"
+	"io"
+)
+
+const HeaderBytesLen uint64 = 32
+
+var (
+	// The fixed prefix of a CAR v2, signalling the version number to previois versoions for graceful fail over.
```
rvagg, line 11:
```suggestion
	// The fixed prefix of a CAR v2, signalling the version number to previous implementations for graceful fail over.
```

masih:
Please note, this PR is closed in favour of a series of smaller PRs with `wip/v2` as base branch.

v2/writer.go
```diff
@@ -0,0 +1,86 @@
...
+	"context"
+	"github.com/ipfs/go-cid"
+	format "github.com/ipfs/go-ipld-format"
+	car_v1 "github.com/ipld/go-car"
+	"github.com/willscott/carbs"
+	"io"
+)
+
+type (
+	// Writer writes CAR v2 into a give io.Writer.
```
rvagg, line 14:
```suggestion
	// Writer writes CAR v2 into a given io.Writer.
```

PR #79, opened 2021-06-10
Define CAR v2 as a separate go module

masih:
Note, the entire approach to CAR v2 implementation and package structure is work in progress, hence this PR is against `wip/v2`. For context see https://github.com/ipld/go-car/pull/78#discussion_r648314538

Since we are releasing a new major version, define a new go module for
the CAR v2 implementation, following the recommendations here:
- https://blog.golang.org/v2-go-modules

The module is versioned as go 1.15 to maintain compatibility as pointed
out by comment here:
- https://github.com/ipld/go-car/pull/75#discussion_r644634850

### DIFF
--- a/v2/doc.go
+++ b/v2/doc.go
@@ -1,0 +1,3 @@
+// Package car represents the CAR v2 implementation.
+// TODO add CAR v2 byte structure here.
+package car
--- a/v2/go.mod
+++ b/v2/go.mod
@@ -1,0 +1,3 @@
+module github.com/ipld/go-car/v2
+
+go 1.15


magik6k (commented):
/v2 may not work too well with how Go handles v2/vX tags

mvdan (approved):
prefix/v1.2.3 is recognized by Go, but I think this should be fine given it's not semver. you could always go for wip-v2 to be on the safer side.

masih:
Thank you both for the reviews. I'll go ahead with merge since all of this is under WIP anyway. once there is a functional mass we can then re-evaluate the module structure for the path of least resistance 🍻

PR #80, opened 2021-06-10
Define CAR v2 fixed prefix and header

masih:
Define the magic CAR v2 prefix proposed in ipld/specs#248 and assert
that it remains a valid CAR v1 header, along with other tests that
verify the expected length, content and version number.

Define basic structs to represent CAR v2 header, along with placeholder
for the future "characteristics" bitfield, and the optional padding
between CAR v1 dump and index added to the end of a CAR v2 archive.

Relates to https://github.com/ipld/specs/pull/248#issuecomment-833141588


v2/car.go
```diff
@@ -0,0 +1,46 @@
...
+
+type (
+	// Header represents the CAR v2 header/pragma.
+	Header struct {
+		// 128-bit characteristics of this CAR v2 file, such as order, deduplication, etc. Reserved for future use.
+		Characteristics Characteristics
+		// The size of CAR v1 encapsulated in this CAR v2 as bytes.
+		CarV1Size uint64
+		// The offset from the beginning of the file at which the CAR v2 index begins.
+		IndexOffset uint64
```
willscott, line 29: can we throw in a CarV1Offset here?

We may find that there's some nice block/cache alignment stuff where it's useful to be able to do things like start the CarV1 at 4KB into the file so that it's block aligned.

v2/car.go
```diff
@@ -0,0 +1,46 @@
+package car
+
+const prefixBytesSize = 16
```
willscott, line 3: why do we have these constant lower case `prefixBytesSize` that are separate from the exported variables?


I think i'd weakly encourage exported, pre-calculated consts so that there's no danger of them being changed

v2/car.go
masih: See https://github.com/ipld/go-car/pull/80/commits/4d78ab2415c674f477a2476f3d6ecb0f62c78549.

rvagg (approved):
nice, I especially like the `TestCarV2PrefixIsValidCarV1Header` in here to tie it in to v1

PR #81, opened 2021-06-12
Car v2 writer using `format.NodeGetter`

masih:
Implement a CAR v2 writer that uses the primitives used by the CAR v1 `Write*` functions for compatibility purposes. Traverse a given list of root CIDs using `format.NodeGetter`, and produce a CAR v2 file, _including the index_, consisting of:
1. **Version prefix** - the prefix readable as a valid Car v1 header, signalling the version as `2`.
2. **CAR v2 header** - consisting of CAR v2 specific metadata, such as size and offset of binary segments within CAR v2 and a placeholder of characteristics bitfield.
3. **CAR v1 dump** - a dump of traversed blocks in CAR v1 format, including the CAR v1 header.
4. **Carbs index** - the index of CAR v1 generated using [Carbs](https://github.com/willscott/carbs).

Additionally, provide the ability to add padding prior to writing 3 (CAR v1 dump) and 4 (Carbs index). This is to provide flexibility when evolving CAR format in future versions as well as optimisation opportunities by the way of offset alignment.

Note, writer test coverage will be improved in future PRs once Carbs implementation is transferred into this repo and refactored to expose the number of written bytes.


v2/car.go
```diff
@@ -17,16 +22,17 @@ var (
...
 	EmptyCharacteristics = new(Characteristics)
 )
 
 type (
 	// Header represents the CAR v2 header/pragma.
 	Header struct {
+		io.WriterTo
 		// 128-bit characteristics of this CAR v2 file, such as order, deduplication, etc. Reserved for future use.
-		Characteristics Characteristics
+		Characteristics *Characteristics
```
willscott, line 35: why would this be a pointer? if you want to be able to cast a memory region into a car header, you'd expect it to be the direct bytes

v2/car.go
```diff
@@ -36,17 +42,100 @@ type (
 	}
 	// Characteristics is a bitfield placeholder for capturing the characteristics of a CAR v2 such as order and determinism.
 	Characteristics struct {
+		io.WriterTo
```
willscott, line 45: this doesn't seem like a property of the characteristics, but rather a method as you have on line 52 below

willscott (commented):
Cool. this should be able to be largely unchanged but more efficient for the `WriterAt` case.

v2/car.go
masih: Though not required, the idea both here and in other places, like [`Header`](https://github.com/ipld/go-car/blob/6ba30fb61f4d815422602659975e28198284e773/v2/car.go#L33) was to document that the interface implemented by the structs purely for human readability. 

This does however add an anonymous field of type `io.Writer` to the structs.
Thank you for the feedback. I will remove it now that it seem to cause confusion and therefore not serving its intended purpose.

v2/car.go
```diff
@@ -1,10 +1,15 @@
...
+	"io"
+)
+
 const (
 	// HeaderBytesSize is the fixed size of CAR v2 header in number of bytes.
-	HeaderBytesSize = 40
+	HeaderBytesSize uint64 = 40
 	// CharacteristicsBytesSize is the fixed size of Characteristics bitfield within CAR v2 header in number of bytes.
-	CharacteristicsBytesSize = 16
+	CharacteristicsBytesSize uint64 = 16
```
mvdan, line 12: any reason to make these typed? afaik there should be none

v2/car.go
```diff
@@ -17,9 +22,9 @@ var (
 		0x02, // uint(2)
 	}
 	// The size of the CAR v2 prefix in 11 bytes, (i.e. 11).
-	PrefixBytesSize = len(PrefixBytes)
+	PrefixBytesSize = uint64(len(PrefixBytes))
 	// Reserved 128 bits space to capture future characteristics of CAR v2 such as order, duplication, etc.
-	EmptyCharacteristics = new(Characteristics)
+	EmptyCharacteristics = Characteristics{0, 0}
```
mvdan, line 27: Anyone can construct this with just `Characteristics{}` or `var zero Characteristics`, right? Seems redundant to expose this. You could always have it unexported if you prefer it for your own code.

v2/car.go
```diff
@@ -17,9 +22,9 @@ var (
 		0x02, // uint(2)
 	}
 	// The size of the CAR v2 prefix in 11 bytes, (i.e. 11).
-	PrefixBytesSize = len(PrefixBytes)
+	PrefixBytesSize = uint64(len(PrefixBytes))
```
mvdan, line 25: I also think this is a bit redundant; `len(PrefixBytes)` is self-explanatory and easy to type. The user can also choose what integer type they want to convert to.

v2/car.go
```diff
@@ -41,12 +46,90 @@ type (
...
+		CarV1Size:       carV1Size,
+	}
+	header.CarV1Offset = PrefixBytesSize + HeaderBytesSize
+	header.IndexOffset = header.CarV1Offset + carV1Size
+	return header
+}
+
+// Size gets the size of Header in number of bytes.
+func (h *Header) Size() uint64 {
+	return HeaderBytesSize
```
mvdan, line 82: I think you should either expose the global constant, or the method, but not both - that seems redundant. Given that the size is entirely static, I'd just remove the method. Otherwise it seems like the Header's size is dependent on some of its fields, given that it's a method.

v2/car.go
```diff
@@ -41,12 +46,90 @@ type (
...
+	wn, err := h.Characteristics.WriteTo(w)
+	if err != nil {
+		return
+	}
+	n += wn
+	wn, err = writeUint64To(w, h.CarV1Offset)
+	if err != nil {
+		return
+	}
+	n += wn
```
mvdan, line 115: nit: you only use `wn` when `err == nil`, and it's always `8`, so I think you could simplify these three uint64 writes to:

```
if err = writeUint64To(w, h.CarV1Offset); err != nil {
    return
}
n += 8
```

At which point `writeUint64To(w, x)` would just be shorthand for `binary.Write(w, binary.LittleEndian, x)`, so it's up to you if it's worth keeping to save a few keystrokes.

v2/car_test.go
```diff
@@ -1,6 +1,7 @@
 package car_test
 
 import (
+	"bytes"
 	cbor "github.com/ipfs/go-ipld-cbor"
 	car_v1 "github.com/ipld/go-car"
 	car_v2 "github.com/ipld/go-car/v2"
```
mvdan, line 7: nit: you should avoid foo_bar names in Go, even for packages; I think carv1 and carv2 would work well here. I'd only resort to foo_bar in very extreme cases, where there's no "single word" one can use, but hopefully that should never happen.

v2/car_test.go
```diff
@@ -74,5 +75,109 @@ func TestEmptyCharacteristics(t *testing.T) {
...
+
+func TestHeader_Marshal(t *testing.T) {
+	tests := []struct {
+		name        string
+		target      car_v2.Header
+		wantMarshal []byte
+		wantErr     bool
+	}{
+		{
+			"header with empty characteristics is marshalled as expected",
```
mvdan, line 89: nit: test and sub-test names as printed by `go test` can't contain spaces, so those will get converted to underscores.

that will also look somewhat inconsistent, like `TestHeader_Marshal/header_with_empty_characteristics_is_marshalled_as_expected`.

IMO that's also a bit confusing, as underscores in places like `TestHeader_Marshal` mean a separation, not simply splitting words.

I personally use camel case for subtest names, so it would look like:

> TestHeader_Marshal/HeaderWithEmptyCharacteristicsIsMarshalledAsExpected

v2/writer.go
```diff
@@ -0,0 +1,125 @@
...
+		IndexPadding Padding
+
+		ctx   context.Context
+		roots []cid.Cid
+	}
+)
+
+// WriteTo writes this padding to the given writer as default value bytes.
+func (p Padding) WriteTo(w io.Writer) (n int64, err error) {
+	paddingBytes := make([]byte, p)
```
mvdan, line 31: I disagree with using `make([]byte, N)` for padding; if I instruct 100MiB of padding because I expect to write lots of blocks, should that really allocate that much at once?

If you're implementing a WriteTo, at least you can be more clever here, such as writing small chunks of zeros until you've written N bytes, via something like `var zero [1024]byte` and a loop.

Or, if you prefer something a bit higher level, you could look at combining `io.CopyN` with a reader that just keeps giving zeros, but I imagine that would end up being more lines anyway.

v2/writer.go
```diff
@@ -0,0 +1,125 @@
...
+
+// NewWriter instantiates a new CAR v2 writer.
+// The writer instantiated uses `carbs.IndexSorted` as the index codec,
+// and `car_v1.DefaultWalkFunc` as the default walk function.
+func NewWriter(ctx context.Context, ng format.NodeGetter, roots []cid.Cid) *Writer {
+	return &Writer{
+		Walk:       car_v1.DefaultWalkFunc,
+		IndexCodec: carbs.IndexSorted,
+		NodeGetter: ng,
+		ctx:        ctx,
```
mvdan, line 45: not too excited about the writer taking a context, but I get that WriterTo taking a context itself would break the io interface.

I guess our use case is weird, because the writing incurs IPLD walking, which should use a context. Most WriterTos don't need a context to write themselves.

I assume we're fine with this given that a Writer is short-lived. Should we make WriteTo error if it's already been called? Using a single Writer to write many times feels wrong, because that would mean reusing the same context, which is almost always not what you want. It also seems wrong to want to marshal exactly the same CAR multiple times in a row: people should marshal it once, and reuse the bytes.

v2/writer.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+// WriteTo writes the given root CIDs according to CAR v2 specification, traversing the DAG using the
+// Writer#Walk function.
+func (w *Writer) WriteTo(writer io.Writer) (n int64, err error) {
+	n, err = w.writePrefix(writer)
+	if err != nil {
+		return
+	}
+	// We read the entire car into memory because carbs#GenerateIndex takes a reader.
```
mvdan, line 57: use a period instead of #, since that's the Go syntax. I assume you used the godoc/pkgsite URL syntax here, but IMO most people expect just Go syntax in docs.

v2/writer.go
```diff
@@ -0,0 +1,125 @@
...
+func (w *Writer) writeHeader(writer io.Writer, carV1Len int) (int64, error) {
+	header := NewHeader(uint64(carV1Len)).
+		WithCarV1Padding(w.CarV1Padding).
+		WithIndexPadding(w.IndexPadding)
+	return header.WriteTo(writer)
+}
+
+func (w *Writer) writePrefix(writer io.Writer) (int64, error) {
+	n, err := writer.Write(PrefixBytes)
+	return int64(n), err
```
mvdan, line 107: I think you can do away with two-line methods like these, which are only called once :)

v2/car.go
```diff
@@ -41,12 +46,90 @@ type (
...
 }
 
 // Size gets the size of Characteristics in number of bytes.
-func (c *Characteristics) Size() int {
+func (c Characteristics) Size() uint64 {
 	return CharacteristicsBytesSize
 }
+
+// NewHeader instantiates a new CAR v2 header, given the byte length of a CAR v1.
+func NewHeader(carV1Size uint64) *Header {
```
mvdan, line 70: I wouldn't make this New func return a pointer, because the Header type is tiny. It barely takes more space than a slice on a 64-bit GOARCH. It's nice to be consistent with always returning pointers from "new" funcs, but it's even nicer to not allocate when we really don't need to :)

v2/writer.go
```diff
@@ -0,0 +1,125 @@
...
+
+	wn, err = w.writeIndex(writer, carV1Bytes)
+	if err == nil {
+		n += wn
+	}
+	return
+}
+
+func (w *Writer) writeHeader(writer io.Writer, carV1Len int) (int64, error) {
+	header := NewHeader(uint64(carV1Len)).
```
mvdan, line 99: this is why I don't want NewHeader to allocate: there's really no reason to allocate here, it should happen on the stack.

v2/car.go
```diff
@@ -41,12 +46,90 @@ type (
...
+// The CAR v1 offset is calculated as the sum of PrefixBytesLen, HeaderBytesLen and the given padding.
+// The call to this function also shifts the Header#IndexOffset forward by the given padding.
+func (h *Header) WithCarV1Padding(padding Padding) *Header {
+	h.CarV1Offset = h.CarV1Offset + uint64(padding)
+	h.IndexOffset = h.IndexOffset + uint64(padding)
+	return h
+}
+
+// WriteTo serializes this header as bytes and writes them using the given io.Writer.
+func (h *Header) WriteTo(w io.Writer) (n int64, err error) {
```
mvdan, line 105: ditto no pointer, just like Characteristics.

v2/writer.go
```diff
@@ -0,0 +1,125 @@
...
+	paddingBytes := make([]byte, p)
+	written, err := w.Write(paddingBytes)
+	n = int64(written)
+	return
+}
+
+// NewWriter instantiates a new CAR v2 writer.
+// The writer instantiated uses `carbs.IndexSorted` as the index codec,
+// and `car_v1.DefaultWalkFunc` as the default walk function.
+func NewWriter(ctx context.Context, ng format.NodeGetter, roots []cid.Cid) *Writer {
```
mvdan, line 40: FWIW I think it is actually fine for Writer to return a pointer, since this is usually what the end user will hold on to, and we might need pointer receiver methods later on to expose more options.

v2/writer.go
```diff
@@ -0,0 +1,125 @@
...
+	return header.WriteTo(writer)
+}
+
+func (w *Writer) writePrefix(writer io.Writer) (int64, error) {
+	n, err := writer.Write(PrefixBytes)
+	return int64(n), err
+}
+
+func (w *Writer) encodeCarV1() (buf *bytes.Buffer, err error) {
+	buf = new(bytes.Buffer)
```
mvdan, line 111: perf nit: you could make the Writer have a bytes.Buffer field to use here, so encodeCarV1 doesn't need to do a second allocation just for a buffer. Remember that the bytes.Buffer is actually tiny, as it only holds the underlying bytes via a []byte.

feel free to keep this as a TODO for now though, especially since you don't have any benchmark.

v2/writer.go
```diff
@@ -0,0 +1,125 @@
...
+	err = car_v1.WriteCarWithWalker(w.ctx, w.NodeGetter, w.roots, buf, w.Walk)
+	return
+}
+
+func (w *Writer) writeIndex(writer io.Writer, carV1 []byte) (n int64, err error) {
+	reader := bytes.NewReader(carV1)
```
mvdan, line 117: hmm, I'd really love a TODO here about removing this extra layering once you incorporate carbs. You essentially do:

1) write v1 into a buffer (which copies)
2) obtain the buffer's bytes
3) set up a Reader atop the buffer (which allocates)
4) have carbs read from the Reader (which copies all the bytes again)

For example, if the carbs API worked on []byte, you'd avoid the steps 3 and 4. Perhaps there are even better ways to avoid the extra copying. I want to make sure we're careful to not copy extra times, given that the inner carv1 could be pretty big.

v2/writer.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/ipfs/go-cid"
+	format "github.com/ipfs/go-ipld-format"
+	car_v1 "github.com/ipld/go-car"
+	"github.com/willscott/carbs"
+	"io"
+)
+
+type (
+	// Padding represents the number of padding bytes.
+	Padding uint64
```
mvdan, line 15: Maybe unexport this; it largely seems like an implementation detail to me. The user should only care about passing you uint64, and then you could have a `writePadding` func that does the logic of writing a certain amount of zero memory.

I also think the option methods would be easier to use with uint64 parameters, as otherwise users might have to explicitly convert to carv2.Padding.

v2/writer.go
mvdan: Another option is to give up io.WriterTo; the user, instead of writing:

    n, err := carv2.NewWriter(ctx, ng, roots).WriteTo(w)

would write something like:

    n, err := carv2.Marshal(ctx, w, ng, roots)

I lean slightly in favor of the second form, because it avoids the anti-pattern of holding a context in a struct field, and avoids possible misuses of the API like holding onto a Writer after the marshaling has completed.

The only possible downside to my mind is that we can't add more parameters or options in the future. For that reason, we could add a `...MarshalOption` as a last parameter, so in the future we could expose marshal option funcs for any possible options we might need.

mvdan (approved):
Looks good overall! I haven't paid *too* much attention to the exported API, we can take a closer look at that via pkg.go.dev once your branch is mostly ready.

v2/car.go
masih: They are used to [calculate offsets](https://github.com/ipld/go-car/blob/b4173dd99357472dd5afb314324db5c96f526425/v2/car.go#L75), the type of which is `uint64`. Making them typed here, and consistently as uint64 avoids the need to cast every time we want to calculate offsets.

v2/car.go
masih: Would it make sense to have it so that we define what empty characteristic means? The rationale is that, we might want default or empty characteristics to have some non-zero defaults, e.g. Ordered, or Dererministic. In which case, having this variable signifies what the characteristics would be if not specified by the user.

v2/car.go
masih: We could but then would it make sense to get the length of a constant byte prefix and cast it to uint64 on every offset calculation?

v2/car.go
masih: I was hoping to introduce a `Sizer` interface or something to that effect for all segments that make up a car. This is to make calculation of a CAR's size to be a series of `Size()` calls on segments. Happy to remove if you think this is redundant.

v2/car.go
masih: The hard-coded 8 in the above approach is what I don't like, since going by encapsulation rules `writeUint64To` does not necessarily have to write 8 bits (even though it does that now) :) 

Having said that, I'll change that to what you suggested, since it is simpler and functionally equal

v2/car_test.go
masih: Thanks for pointing that out. That's a reasonable approach. 

My take is that CamelCase is harder to read than simply spaces for humans whom look at the code. But then it keeps the test name and `go test` output consistent.

The `_` in the test method name signifies a receiver function defined for type `Header`.  I used to call the function `Marshal`; It is now called `WriteTo` so I'll fix that. 

Is the CamelCase the convention generally followed in IPLD? Happy to name tests in that convention if that keeps things consistent.

v2/writer.go
masih: Sounds good thanks @mvdan I am with you on ditching the `io.Writer` I think it's only a matter of time for us to have to break it at some point.

v2/car.go
mvdan: I'm pretty sure that expression would work with untyped constants. They get converted to whatever type is necessary, implicitly.

masih: ah right you are; thank you for pointing this out

v2/car.go
mvdan: I'm not sure. I think I'd expect "all zero bits" to be the default forever, but I'm also not sure what kind of use cases we have in mind for characteristics in the future. You could later on change the writer to set some of the bits by default, and I wonder whether you could say that's fully backwards compatible - it could mean that some downstreams would suddenly start writing different CARs by just upgrading the library, which could definitely surprise developers.

v2/car.go
mvdan: If you prefer to have this variable for your own implementation, I think that's fine, but I don't think you need to expose it publicly. Anything that gets made public has to stay there forever, and makes the godoc index larger.

v2/car.go
mvdan: If you prefer an interface-like system, that's also an option, but then I'd unexport the constants. You really don't want to expose multiple ways to do the same thing in the documentation unless strictly necessary, because that can confuse users and make it harder to start using the library.

I think I'd still lean on constants, because obtaining the header size via `Header{}.Size()` feels slightly roundabout. I also can't think of any advantage to the interface other than "it looks nice" :)

v2/car.go
mvdan: If you think you'll write integers of different lengths, by all means do it some other way - I'm just working from what I see in the current code. Worth noting that I'm pretty sure cbor allows for shorter integers than 64-bit, so I wonder if the carv2 draft spec specifically says we should use 64 bits for all of these.

v2/car_test.go
mvdan: Right, consistency is what I'm after here. There's the consistency between the code and the output, but also the consistency that "_" is often a separator between type and method, or a separator between a name and a description, like `ExampleWriter_veryLarge`.

I don't think there's a hard convention anywhere about this, actually. The official docs happen to not use multiple words, only subtest names like Test1. And blog posts like https://blog.golang.org/subtests actually use a mix of both, which I find pretty funny.

So I guess follow what's consistent with the rest of go-car :) I still lean towards camel case by default, but I can't say it's the one true way.

v2/writer.go
masih: There are `WithPadding` methods in the Header that use this type. Do you think we should get those to use some other type if we unexpert this?

mvdan: Yes; those could just take uint64, so I don't see a reason to do anything else.

v2/car.go
masih: Good point thanks

v2/writer.go
mvdan: Another reason to keep this a TODO: once you incorporate carbs, maybe you'll do something more clever than using a bytes.Buffer.

v2/car_test.go
masih: There is no current strong convention in go-car. Let's make the CamelCase it :) Thanks @mvdan

v2/writer.go
masih: @mvdan if you don't mind I address this in a sparate PR?

v2/car.go
```diff
@@ -41,12 +45,70 @@ type (
...
+	header.IndexOffset = header.CarV1Offset + carV1Size
+	return header
+}
+
+// WithIndexPadding sets the index offset from the beginning of the file for this header and returns the
+// header for convenient chained calls.
+// The index offset is calculated as the sum of PrefixBytesLen, HeaderBytesLen,
+// Header.CarV1Len, and the given padding.
+func (h Header) WithIndexPadding(padding uint64) Header {
+	h.IndexOffset = h.IndexOffset + padding
```
rvagg, line 76: @masih when we do this, do we lose the ability to tell exactly where the CARv1 ends? I think we need to retain the ability to derive `carV1Size` so we can parse it successfully using the CARv1 parser, which only knows "end" when the reader runs out of bytes. If we feed it zero-padding at the end we're going to have problems I think--not just parsing problems either, if we continue to use exact CARv1 bytes to generate CommP for Filecoin then we need to know the exact bytes.

v2/writer.go
```diff
@@ -0,0 +1,138 @@
...
+// The writer instantiated uses `carbs.IndexSorted` as the index codec,
+// and `carv1.DefaultWalkFunc` as the default walk function.
+func NewWriter(ctx context.Context, ng format.NodeGetter, roots []cid.Cid) *Writer {
+	return &Writer{
+		Walk:         carv1.DefaultWalkFunc,
+		IndexCodec:   carbs.IndexSorted,
+		NodeGetter:   ng,
+		ctx:          ctx,
+		roots:        roots,
+		encodedCarV1: new(bytes.Buffer),
```
rvagg, line 66: @masih I'm assuming that this is an initial, easy version and that buffering the entirety of CARv1 in memory before writing isn't something we're planning on being stuck with? One of the nice features of using fixed byte length blocks at the start (as opposed to the usual varint length blocks that we do most other places around our stack) is the ability to come back and update the header details _after_ writing CARv1 bytes.

v2/car.go
rvagg: fwiw re scale, Filecoin storage deals can be packed into ~30Gb CAR files, which is a common case we're going to need to be OK with (double that for miners doing 64G sectors). The [full chain snapshot for Filecoin](https://docs.filecoin.io/get-started/lotus/chain/#full-chain-snapshot) is probably the biggest CAR around, I think it's in the 420Gb range. The latter is probably not a CARv2 candidate, but the former is, at least on the miner side. So we're probably going to want a clear route to non-buffering writes.

v2/writer.go
masih: Absolutely; all the copied car content will go away in the coming PRs. I have left a bunch of TODOs in code; it depends on carbs API refactoring in places.

v2/car.go
masih: > when we do this, do we lose the ability to tell exactly where the CARv1 ends?

The header contains a dedicated [field for storing the size of car v1 dump](https://github.com/masih/go-car/blob/ad8a17eaac3171c30f9bfcadc6464bd8e5c96a3b/v2/car.go#L37). Did I miss something?

> a clear route to non-buffering writes

Of course; all of this is living in a WIP branch. No buffering of car content in memory for the final version :)

rvagg: > The header contains a dedicated field for storing the size of car v1 dump. Did I miss something?

nope, I did! all good then.

PR #82, opened 2021-06-15
Copy `carbs` into `go-car` repository

masih:
[carbs](https://github.com/willscott/carbs) implements indexing of CAR v1, and is used by CAR v2. Copy the code straight-up with minimal modifications from its original source. Future PRs would then refactor it in small iterations. The only modifications made here are the ones to satisfy `staticcheck`.


v2/carbs/carbs.go
```diff
@@ -0,0 +1,300 @@
...
+	"bufio"
+	"bytes"
+	"context"
+	"encoding/binary"
+	"fmt"
+	"io"
+
+	blocks "github.com/ipfs/go-block-format"
+	"github.com/ipfs/go-cid"
+	bs "github.com/ipfs/go-ipfs-blockstore"
```
mvdan, line 13: I'd use a slightly better name here, like blockstore; a global "bs" name is too short to be intuitive

v2/carbs/carbs.go
```diff
@@ -0,0 +1,300 @@
...
+	"github.com/multiformats/go-multihash"
+
+	pb "github.com/cheggaaa/pb/v3"
+	car "github.com/ipld/go-car"
```
mvdan, line 17: carv1, for consistency with the rest of this repo?

v2/carbs/carbs.go
```diff
@@ -0,0 +1,300 @@
...
+
+// Get gets a block from the store
+func (c *Carbs) Get(key cid.Cid) (blocks.Block, error) {
+	offset, err := c.idx.Get(key)
+	if err != nil {
+		return nil, err
+	}
+	entry, bytes, err := c.Read(int64(offset))
+	if err != nil {
+		fmt.Printf("failed get %d:%v\n", offset, err)
```
mvdan, line 110: I think this deserves a TODO; a library should never print directly to stdout/stderr

v2/carbs/carbs.go
```diff
@@ -0,0 +1,300 @@
...
+	pb "github.com/cheggaaa/pb/v3"
+	car "github.com/ipld/go-car"
+	"github.com/ipld/go-car/util"
+	"golang.org/x/exp/mmap"
+)
+
+var errNotFound = bs.ErrNotFound
+
+// Carbs provides a read-only Car Block Store.
+type Carbs struct {
```
mvdan, line 25: nit: I think `carbs.BlockStore` would be more intuitive and less repetitive than `carbs.Carbs`, and also more consistent with `carbs.Index`

v2/carbs/carbs_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+func TestIndexRT(t *testing.T) {
+	/*
+		carFile, err := mkCar()
+		if err != nil {
+			t.Fatal(err)
+		}
+		defer os.Remove(carFile)
+	*/
+	carFile := "test.car"
```
mvdan, line 54: I think this will edit files in the package directory where the Go files are, meaning it could mess with the source code or leave files behind if the test panics/fails too early.

I'd use t.TempDir in all these that work with paths on the filesystem - feel free to leave as a TODO.

mvdan: Also, that file should be put in `v2/carbs/testdata/test.car`, because otherwise the Go module zip will include the file as part of the `v2/carbs` package source code

v2/carbs/util/hydrate.go
```diff
@@ -0,0 +1,49 @@
+package main
+
+import (
+	"fmt"
+	"github.com/ipld/go-car/v2/carbs"
```
mvdan, line 5: nit: keep the first group of imports to just the standard library. you can use gofumpt if you want to automate that :)

mvdan (approved):
Just minor nits on how the code would fit better into the go-car repo, I'm not reviewing Will's code too closely.

PR #84, opened 2021-06-16
Implement CAR v2 Reader

masih:
Provide an API to read sections of a CAR v2 as well as a utility to get a `Blockstore` from a CAR v2.

### DIFF
--- a/v2/car.go
+++ b/v2/car.go
@@ -6,6 +6,8 @@ import (
 )
 
 const (
+	// The size of the CAR v2 prefix in 11 bytes, (i.e. 11).
+	PrefixBytesSize = 11
 	// HeaderBytesSize is the fixed size of CAR v2 header in number of bytes.
 	HeaderBytesSize = 40
 	// CharacteristicsBytesSize is the fixed size of Characteristics bitfield within CAR v2 header in number of bytes.
@@ -22,8 +24,6 @@ var (
 		0x76, 0x65, 0x72, 0x73, 0x69, 0x6f, 0x6e, // "version"
 		0x02, // uint(2)
 	}
-	// The size of the CAR v2 prefix in 11 bytes, (i.e. 11).
-	prefixBytesSize = uint64(len(PrefixBytes))
 )
 
 type (
@@ -47,11 +47,23 @@ type (
 
 // WriteTo writes this characteristics to the given writer.
 func (c Characteristics) WriteTo(w io.Writer) (n int64, err error) {
-	if err = writeUint64To(w, c.Hi); err != nil {
+	if err = binary.Write(w, binary.LittleEndian, c.Hi); err != nil {
 		return
 	}
 	n += uint64BytesSize
-	if err = writeUint64To(w, c.Lo); err != nil {
+	if err = binary.Write(w, binary.LittleEndian, c.Lo); err != nil {
+		return
+	}
+	n += uint64BytesSize
+	return
+}
+
+func (c Characteristics) ReadFrom(r io.Reader) (n int64, err error) {
+	if err = binary.Read(r, binary.LittleEndian, &c.Hi); err != nil {
+		return
+	}
+	n += uint64BytesSize
+	if err = binary.Read(r, binary.LittleEndian, &c.Lo); err != nil {
 		return
 	}
 	n += uint64BytesSize
@@ -63,7 +75,7 @@ func NewHeader(carV1Size uint64) Header {
 	header := Header{
 		CarV1Size: carV1Size,
 	}
-	header.CarV1Offset = prefixBytesSize + HeaderBytesSize
+	header.CarV1Offset = PrefixBytesSize + HeaderBytesSize
 	header.IndexOffset = header.CarV1Offset + carV1Size
 	return header
 }
@@ -94,21 +106,38 @@ func (h Header) WriteTo(w io.Writer) (n int64, err error) {
 		return
 	}
 	n += wn
-	if err = writeUint64To(w, h.CarV1Offset); err != nil {
+	if err = binary.Write(w, binary.LittleEndian, h.CarV1Offset); err != nil {
 		return
 	}
 	n += uint64BytesSize
-	if err = writeUint64To(w, h.CarV1Size); err != nil {
+	if err = binary.Write(w, binary.LittleEndian, h.CarV1Size); err != nil {
 		return
 	}
 	n += uint64BytesSize
-	if err = writeUint64To(w, h.IndexOffset); err != nil {
+	if err = binary.Write(w, binary.LittleEndian, h.IndexOffset); err != nil {
 		return
 	}
 	n += uint64BytesSize
 	return
 }
 
-func writeUint64To(w io.Writer, v uint64) error {
-	return binary.Write(w, binary.LittleEndian, v)
+func (h Header) ReadFrom(r io.Reader) (n int64, err error) {
+	rn, err := h.Characteristics.ReadFrom(r)
+	if err != nil {
+		return
+	}
+	n += rn
+	if err = binary.Read(r, binary.LittleEndian, &h.CarV1Offset); err != nil {
+		return
+	}
+	n += uint64BytesSize
+	if err = binary.Read(r, binary.LittleEndian, &h.CarV1Size); err != nil {
+		return
+	}
+	n += uint64BytesSize
+	if err = binary.Read(r, binary.LittleEndian, &h.IndexOffset); err != nil {
+		return
+	}
+	n += uint64BytesSize
+	return
 }
--- a/v2/car.go
+++ b/v2/car.go
@@ -6,13 +6,13 @@ import (
 )
 
 const (
-	// The size of the CAR v2 prefix in 11 bytes, (i.e. 11).
-	PrefixBytesSize = 11
-	// HeaderBytesSize is the fixed size of CAR v2 header in number of bytes.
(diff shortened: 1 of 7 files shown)


v2/car.go
```diff
@@ -6,6 +6,8 @@ import (
 )
 
 const (
+	// The size of the CAR v2 prefix in 11 bytes, (i.e. 11).
+	PrefixBytesSize = 11
```
mvdan, line 10: nit: I think "BytesSize" could be replaced by just "Size" in all these.

v2/car.go
```diff
@@ -47,11 +47,23 @@ type (
...
-	if err = writeUint64To(w, c.Hi); err != nil {
+	if err = binary.Write(w, binary.LittleEndian, c.Hi); err != nil {
+		return
+	}
+	n += uint64BytesSize
+	if err = binary.Write(w, binary.LittleEndian, c.Lo); err != nil {
+		return
+	}
+	n += uint64BytesSize
+	return
```
mvdan, line 58: Once you have more than a handful of lines I'd personally avoid naked returns: https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters

They're not wrong per se, but once a function goes past a trivial line length, it requires more work from the reader to remember what the implicit return parameters are (or if there were any!) and how they were assigned to.

v2/car.go
```diff
@@ -94,21 +106,38 @@ func (h Header) WriteTo(w io.Writer) (n int64, err error) {
...
+	if err = binary.Write(w, binary.LittleEndian, h.IndexOffset); err != nil {
 		return
 	}
 	n += uint64BytesSize
 	return
 }
 
-func writeUint64To(w io.Writer, v uint64) error {
-	return binary.Write(w, binary.LittleEndian, v)
+func (h Header) ReadFrom(r io.Reader) (n int64, err error) {
```
mvdan, line 124: once you start benchmarking all this, you will probably end up finding that reading the entire header as a single chunk of []byte would most likely be faster. It avoids extra Read calls, and allows working directly with slices of byte. Plus, the header is pretty small in terms of byte buffers, anyway.

v2/car.go
```diff
@@ -94,21 +106,38 @@ func (h Header) WriteTo(w io.Writer) (n int64, err error) {
...
+	rn, err := h.Characteristics.ReadFrom(r)
+	if err != nil {
+		return
+	}
+	n += rn
+	if err = binary.Read(r, binary.LittleEndian, &h.CarV1Offset); err != nil {
```
mvdan, line 130: perf nit, for once you start benchmarking: binary.Read also uses a type switch on the last parameter, which is an interface, so you do quite a bit of branching for every integer you parse. it will likely be faster to use binary.LittleEndian.Uint64 once you read the entire header as a []byte.

v2/car.go
```diff
@@ -94,21 +106,38 @@ func (h Header) WriteTo(w io.Writer) (n int64, err error) {
...
 		return
 	}
 	n += uint64BytesSize
-	if err = writeUint64To(w, h.CarV1Size); err != nil {
+	if err = binary.Write(w, binary.LittleEndian, h.CarV1Size); err != nil {
 		return
 	}
 	n += uint64BytesSize
-	if err = writeUint64To(w, h.IndexOffset); err != nil {
+	if err = binary.Write(w, binary.LittleEndian, h.IndexOffset); err != nil {
```
mvdan, line 117: it didn't occur to me while reviewing the write side, but it will also likely be significantly cheaper to do all this on []byte, since the sizes here are fixed and pretty small. fine to leave as a TODO until you have benchmarks though, as said before.

v2/offset_reader.go
```diff
@@ -0,0 +1,59 @@
...
+// OffsetReader implements Read, and ReadAt on a section
+// of an underlying io.ReaderAt.
+// The main difference between io.SectionReader and OffsetReader is that
+// NewOffsetReader accepts zero as n, the number of bytes to read, with
+// the trade-off that it does not implement Seek.
+// When n is set to zero, it will delegate io.EOF errors to the underlying
+// io.ReaderAt.
+// This is useful when reading a section at the end of a io.ReaderAt without
+// having to know the total number readable of bytes.
+type OffsetReader struct {
```
mvdan, line 14: I thought we just needed this to implement io.ReaderAt, not also io.Reader?

If we were to propose it for inclusion in the io package it would likely need both, but I think here we just need one method.

also, you could assert it, like:

    var _ io.ReaderAt = (*OffsetReader)(nil)

v2/offset_reader.go
```diff
@@ -0,0 +1,59 @@
...
+	base    int64
+	off     int64
+	limit   int64
+	limited bool
+}
+
+// NewOffsetReader returns an OffsetReader that reads from r
+// starting at offset off and stops with io.EOF after n bytes.
+// If n is set to 0 then it will carry on reading until r reaches io.EOF.
+func NewOffsetReader(r io.ReaderAt, off int64, n int64) *OffsetReader {
```
mvdan, line 25: not sure I understand the point of `n int64`, since it's always going to be 0. if it were > 0, then one could just use io.SectionReader, setting the limit to off+n.

so I think you should remove that parameter, and remove the `limited` boolean as it's always true.

v2/offset_reader.go
```diff
@@ -0,0 +1,59 @@
...
+	return &OffsetReader{r, off, off, off + n, n == 0}
+}
+
+func (o *OffsetReader) Read(p []byte) (n int, err error) {
+	if o.limited {
+		if o.off >= o.limit {
+			return 0, io.EOF
```
mvdan, line 32: I assume we don't actually call Read anywhere, because this seems wrong: you use n==0, so limited==true, and off>=limit will always be true - this will just immediately return EOF no matter what.

what this should do, instead, is:

1) Have I returned an error before, like EOF? If so, return it again. If not, continue.
2) Do a ReadAt at offset. If it errors, store that and return it. Note that you can get EOF at the same time as the last chunk of bytes.
3) Increment the offset and return the bytes.

v2/offset_reader.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+	n, err = o.r.ReadAt(p, o.off)
+	o.off += int64(n)
+	return
+}
+
+func (o *OffsetReader) ReadAt(p []byte, off int64) (n int, err error) {
+	if o.limited {
+		if off < 0 || off >= o.limit-o.base {
+			return 0, io.EOF
```
mvdan, line 46: similar to before, this seems wrong to me, given that you by definition can't know what the limit is. if you knew, you'd just use io.SectionReader.

v2/reader.go
```diff
@@ -0,0 +1,93 @@
...
+	return
+}
+
+func (r *Reader) headerReader() *io.SectionReader {
+	return io.NewSectionReader(r.r, PrefixBytesSize, HeaderBytesSize)
+}
+
+func (r *Reader) CarV1() (*carv1.CarReader, error) {
+	carV1Section := r.carV1Reader()
+	return carv1.NewCarReader(bufio.NewReader(carV1Section))
```
mvdan, line 66: why put an io.SectionReader through a bufio.Reader? if it's necessary, I'd document that briefly

v2/reader.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+func (r *Reader) carV1Reader() *io.SectionReader {
+	return io.NewSectionReader(r.r, int64(r.Header.CarV1Offset), int64(r.Header.CarV1Size))
+}
+
+func (r *Reader) Index() (carbs.Index, error) {
```
mvdan, line 73: I'm a bit confused by the name Reader and its methods. As a Reader, constructing it doesn't actually do any of the unmarshaling yet, that's fine. Some of the methods like CarV1 also return readers, but others like Index presumably fully parse the index. If that is by design, maybe make it obvious by calling methods that return readers like CarV1Reader?

mvdan (commented):
Have to go AFK soon so here's a first pass.

v2/car.go
mvdan: As in, I don't think anyone would expect another "size" here.

v2/reader.go
masih: Typo; my bad. thanks

v2/offset_reader.go
masih: Do you think it's worth removing the `io.Reader` implementation considering its small size and the fact that it is most likely needed when we propose this for inclusion in io package, despite its under-utilisation here?

mvdan: Up to you :)

v2/car.go
masih: TODO added. thank you for pointing this out

v2/car.go
masih: Will get this resolved when optimising Write in future PRs; thank you for pointing this out 🍻

v2/reader.go
```diff
@@ -0,0 +1,61 @@
...
+	if err := cr.readPrefix(); err != nil {
+		return nil, err
+	}
+	if err := cr.readHeader(); err != nil {
+		return nil, err
+	}
+	return cr, nil
+}
+
+func (r *Reader) readPrefix() (err error) {
```
rvagg, line 35: oo, this is a nice touch, nicely forward compatible and involves carv1 :ok_hand:

rvagg (approved):
very nice

PR #85, opened 2021-06-16
Move `carbon` implementation to `go-car` repo

masih:
[`carbon`](https://github.com/willscott/carbon), originally implemented by @willscott will be heavily used by the CAR v2. Copy the code here so that all CAR v2 related code is captured under `go-car` repo.


masih:
> If you want to keep history, there is a git option for that https://stackoverflow.com/a/10548919

Great suggestion thank you; I'd love to keep the history. I'll do the same for carbs too after this, since it was copied without the history before.

masih:
`carbs` move (2.0) with history is here: https://github.com/ipld/go-car/pull/86

mvdan (approved):
There's also the question of whether these packages should be e.g. v2/internal/carbon, to prevent external users from consuming them directly.

Maybe some people will find them useful. But also, once v2.0.0 ships with exposed APIs, we can't remove or change those anymore.

masih:
> There's also the question of whether these packages should be e.g. v2/internal/carbon, to prevent external users from consuming them directly.
> 
> Maybe some people will find them useful. But also, once v2.0.0 ships with exposed APIs, we can't remove or change those anymore.

I agree. at the very least the packages will be renamed if not combined. All of this is WIP. I _think_ internal makes sense. What folks need is high-level APIs that provide what carbon/carbs do, not necessarily the carbon and carbs API :)

PR #88, opened 2021-06-18
Implement initial read-only blockstore

masih:
* Refactor packages into internal as needed, and implement a `ReadOnlyBlockstore` as provided by `carbs`.
* Improve write efficiency for CAR v2 primitives.

Note, future PRs will adjust the read-only blockstore to accept CAR v2 format and work transparently.


v2/blockstore/ro_blockstore.go
```diff
@@ -0,0 +1,190 @@
...
+var _ blockstore.Blockstore = (*ReadOnlyBlockStore)(nil)
+
+// errUnsupported is returned for unsupported operations
+var errUnsupported = errors.New("unsupported operation")
+
+type (
+	// ReadOnlyBlockStore provides a read-only Car Block Store.
+	ReadOnlyBlockStore struct {
+		backing io.ReaderAt
+		idx     index.Index
```
willscott, line 30: we probably want a non-internal Index type - it sounds like the sharded blockstore would like to be able to detach indexes to store them separate from cars, and also be able to recombine a carv1 + index into a carv2

v2/blockstore/ro_blockstore.go
```diff
@@ -0,0 +1,190 @@
...
+	return errUnsupported
+}
+
+// PutMany is not supported and always returns an error
+func (b *ReadOnlyBlockStore) PutMany([]blocks.Block) error {
+	return errUnsupported
+}
+
+// AllKeysChan returns the list of keys in the store
+func (b *ReadOnlyBlockStore) AllKeysChan(ctx context.Context) (<-chan cid.Cid, error) {
```
willscott, line 141: we may use this walk for populating the index, and we need to be able to iterate keys in this way somewhere for index generation. In general though, when it's asked for all keys from a blockstore with an index, we should iterate through the index when possible rather than linear reads through the full car.

v2/blockstore/ro_blockstore.go
```diff
@@ -0,0 +1,190 @@
+package blockstore
```
mvdan, line 1: this package is called blockstore, so I think blockstore/readonly.go is slightly better than blockstore/ro_blockstore.go

v2/blockstore/ro_blockstore.go
```diff
@@ -0,0 +1,190 @@
...
+
+import (
+	"bufio"
+	"context"
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"golang.org/x/exp/mmap"
```
mvdan, line 9: nit: not a std import, shouldn't be in this group

v2/blockstore/ro_blockstore.go
```diff
@@ -0,0 +1,190 @@
...
+	"github.com/ipld/go-car/v2/internal/index"
+	internalio "github.com/ipld/go-car/v2/internal/io"
+)
+
+var _ blockstore.Blockstore = (*ReadOnlyBlockStore)(nil)
+
+// errUnsupported is returned for unsupported operations
+var errUnsupported = errors.New("unsupported operation")
+
+type (
```
mvdan, line 26: any particular reason you use `()` groups for single types? consistency? it does add two lines for very little benefit, so I'd probably avoid it.

v2/internal/carbon/carbon_fds.go
```diff
@@ -14,8 +15,8 @@ import (
 type carbonFD struct {
 	path        string
 	writeHandle *poswriter
-	carbs.BlockStore
-	idx *insertionIndex
+	blockstore.ReadOnlyBlockStore
```
mvdan, line 18: like the filename, I wonder if this type stutters a bit. `blockstore.ReadOnly` is pretty good, or if you still prefer to clarify what it is, maybe `blockstore.ReadOnlyStore`?

also, nit: I'm pretty sure the go-ipfs-blockstore package uses the name Blockstore, not BlockStore (note the lack of capitalization of Store), so I think go-car should be consistent too. Maybe its name is a historical typo/mistake, but best to not confuse things further with multiple spellings.

v2/writer.go
```diff
@@ -3,12 +3,12 @@ package car
 import (
 	"bytes"
 	"context"
+	"github.com/ipld/go-car/v2/internal/index"
```
mvdan, line 6: hmmm :) maybe run goimports or gofumpt on all the files to make sure you're not mixing things into the std group?

mvdan (approved):

v2/blockstore/ro_blockstore.go
masih: face of habit I usually do this to group type declarations

v2/blockstore/ro_blockstore.go
masih: https://github.com/ipld/go-car/issues/89

v2/blockstore/ro_blockstore.go
masih: Added TODO thank you for pointing this out

rvagg:
The inclusion of `ReadCid` here is unfortunate, it really should be something go-cid can do. We added a similar one in JS recently, `decodeFirst()` and having the "find a CID in these bytes, starting from here" functionality has generic utility.
Not suggesting this is critical to address now, but is there a good reason anyone can see why this shouldn't be on the future-refactor list? (Perhaps something I might get to on a day I need a distraction).

rvagg:
noted here https://github.com/ipfs/go-cid/issues/126

masih:
That makes sense thanks for pointing this out @rvagg 🍻  I have captured [this](https://github.com/ipld/go-car/issues/91) on `go-car` repo so we don't forget to do it

Issue #83, opened 2021-06-15
CAR v2 characteristics should indicate if index is included


masih:
We want to be lazy about generating an index. This is so that if we've already got the carv2 header at the beginning of the car, we can write an index suffix without relocating the actual car data on disk.

rvagg:
I can't recall if/where we discussed this but I think having an "index offset" in the header of `0` should do this for us without having to use the characteristics bitfield. It can be overwritten later if needed since it's in a fixed location and `0` is a good indicator of "not present".

masih:
Agreed, thanks @rvagg  The body of this issue was copied from slack conversations with folks. Closing.

PR #90, opened 2021-06-19
Implement CAR introspector API

masih:
* Implement an API that allows a car file to be introspected, which would
worth for both CAR v1 and v2. When the given reader is a v1, the
`HasIndex` will always be false, and the car v1 size is set to the total
readable bytes in the given reader. When the given reader is a v2, then
the bytes are parsed and corresponding values are fetched from v2
header. Regardless of car version, `Roots` filed is populated, either
from the v1 header, or the v1 dump in car v2.

* Add `HasIndex` placeholder in v2 header characteristics. Note the
placeholder always returns true, since the current writer always writes
the index. When the writer is made configurable we then change the
placeholder to do its thing.

* Add lazy loading of roots in CAR v2 reader, used by the introspector.

* Add utility io function to size the number of readable bytes in a
reader.

Relates to #83


v2/car.go
```diff
@@ -44,6 +44,13 @@ type (
 	}
 )
 
+// HasIndex indicates whether the index is present.
+func (c Characteristics) HasIndex() bool {
+	// Always true because in current implementation we always write the index.
```
willscott, line 49: I think we don't need to have this be a bit set in characteristics, and can instead use [`IndexOffset`](https://github.com/ipld/go-car/blob/cdf53fef4046264042b2fbb089de1ef1ae9e15c1/v2/car.go#L38) being non-zero and valid as an indicator of an index being present.

v2/introspector.go
```diff
@@ -0,0 +1,68 @@
...
+	CarV1Size uint64
+	// The offset at which the index starts if it is present, otherwise 0.
+	IndexOffset uint64
+}
+
+// Introspect introspects the given readable bytes and provides metadata about the characteristics
+// and the version of CAR that r represents regardless of its version. This function is backward
+// compatible; it supports both CAR v1 and v2.
+// Returns error if r does not contain a valid CAR payload.
+func Introspect(r io.ReaderAt) (*Introspection, error) {
```
willscott, line 35: This seems like it will be the primary `Read` operation we'll want clients to make, and is the generalized version of `ReadHeader` that previously existed. (where after this generalized header read you can choose to iterate through the items in the car to import them, or open it as a blockstore).

As such, i wonder if we want to call this `Introspect` versus `ReadHeader` or `Parse` or similar?

v2/internal/io/size.go
```diff
@@ -0,0 +1,19 @@
+package io
+
+import "io"
+
+// Size returns the total number of readable bytes in r, discarding any bytes read.
+func Size(r io.Reader) (uint64, error) {
```
mvdan, line 6: I don't think you need this; you could just use `io.Copy` with `io.Discard`, which gives you a total `n` of bytes copied.

v2/introspector.go
```diff
@@ -0,0 +1,68 @@
...
+	"bufio"
+	"fmt"
+	"github.com/ipfs/go-cid"
+	carv1 "github.com/ipld/go-car"
+	internalio "github.com/ipld/go-car/v2/internal/io"
+	"io"
+)
+
+// Introspection captures the result of an Introspect call.
+type Introspection struct {
```
mvdan, line 13: isn't this very similar to the header type?

v2/reader.go
```diff
@@ -45,6 +47,19 @@ func (r *Reader) readPrefix() (err error) {
 	return
 }
 
+// Roots returns the root CIDs of this CAR
+func (r *Reader) Roots() ([]cid.Cid, error) {
+	if r.roots != nil {
+		return r.roots, nil
+	}
+	header, err := carv1.ReadHeader(bufio.NewReader(r.carv1SectionReader()))
```
mvdan, line 55: weird that v1 wants a bufio.Reader...

v2/car.go
rvagg: sounds reasonable, I'll try and remember to note this in a spec doc

v2/introspector.go
```diff
@@ -0,0 +1,68 @@
...
+	// The version, either 1 or 2.
+	Version uint64
+	// The root CIDs if any.
+	Roots []cid.Cid
+	// Indicates whether the index is present.
+	HasIndex bool
```
rvagg, line 19: As per @willscott's comment above - maybe this isn't necessary and `IndexOffset==0` would be enough? Introducing the possibility of `IndexOffset=0 && HasIndex=true` could be opening awkward doors (I can hear @warpfork gently whispering _"cardinality my friends, it's all about cardinality ..."_).

v2/internal/io/size.go
masih: aha thank you 🍻

v2/introspector.go
masih: That's the thing; which header type? If `CarHeader` in v1 then that doesn't have offsets, if `Header` in v2 then that doesn't have roots nor the version. Not sure what changes would come in v3 but the chances are it would have stuff that are only special to v3. How would we unify them all I asked myself?

I thought a dedicated `Introspection` type could tell us things we want to know about _a_ CAR regardless of its version.

v2/introspector.go
masih: I think `Parse` makes sense; any thoughts on the `Introspection` type name if we call this `Parse`?

`ReadHeader` makes sense in car v1; the issue for v2, however, is that we technically have two headers: the version header (magic prefix that is a valid car v1 header) and the car v2 header. Both these sections have information that a user would want to know as part of introspecting a CAR without knowing its version in advance: 
* version, roots from v1 header, and
* characteristics, offsets from v2 header.

Based on this train of thought I am not sure if `ReadHeader` for introspecting a CAR file that _could be_ v1 or v2 makes sense.

v2/introspector.go
mvdan: Ah, I see. Maybe embed v2's header struct type, and add the extra fields to the parent like version and roots? That way, at least, you don't duplicate the header fields.

I don't think you need to worry about forwards compatibility with v3. If/when that comes, I seem to understand that many pieces of the core design would change, so it's likely that we would need a carv3 module anyway.

masih: OK that makes sense. I embedded `carv1.CarHeader` into `Introspect` too while at it.

v2/introspector.go
mvdan: ReadHeaders? I agree the name "header" is somewhat overloaded in carv2.

If you prefer to avoid the name "header" altogether, to prevent confusion, I think it's fine to have an API called something else like "inspect" or "introspect".

v2/introspector.go
mvdan: Beware that embedding multiple structs can get ugly, e.g. if any names conflict they will not be promoted.

masih: Do you think we should only embed car v2 header?

mvdan: This is a carv2 module, so my intuition would be that APIs should generally be geared towards the v2 header, since the v2 spec is generally a superset of the v1 spec :) I ultimately defer to you on this though, as I haven't thought hard about this API. Maybe a flat struct without embedding is best in your opinion - I just thought that duplicating many of the fields looked a bit off.

v2/reader.go
```diff
@@ -54,10 +70,29 @@ func (r *Reader) readHeader() (err error) {
...
 // IndexReaderAt provides an io.ReaderAt containing the carbs.Index of this CAR v2.
 func (r *Reader) IndexReaderAt() io.ReaderAt {
 	return internalio.NewOffsetReader(r.r, int64(r.Header.IndexOffset))
 }
+
+// ReadPragma reads the pragma from r.
+// This function accepts both CAR v1 and v2 payloads.
+// The roots are returned only if the version of pragma equals 1, otherwise returns nil as roots.
+func ReadPragma(r io.Reader) (version uint64, roots []cid.Cid, err error) {
+	header, err := carv1.ReadHeader(bufio.NewReader(r))
```
mvdan, line 89: A TODO for you: if I provide an io.Reader that already has all the fancy reader methods needed, you should use it directly rather than wrapping the Read method with something else like a bufio.Reader.

In this particular case, if I give you a bufio.Reader, you should use it directly. I imagine there are other cases like this in the carv2 API.

v2/reader.go
```diff
@@ -54,10 +70,29 @@ func (r *Reader) readHeader() (err error) {
...
 }
 
 // IndexReaderAt provides an io.ReaderAt containing the carbs.Index of this CAR v2.
 func (r *Reader) IndexReaderAt() io.ReaderAt {
 	return internalio.NewOffsetReader(r.r, int64(r.Header.IndexOffset))
 }
+
+// ReadPragma reads the pragma from r.
+// This function accepts both CAR v1 and v2 payloads.
+// The roots are returned only if the version of pragma equals 1, otherwise returns nil as roots.
```
mvdan, line 87: It makes me slightly uneasy that this API is *less* powerful with the newer and better CARv2.

Presumably we could always provide the roots, if for a version 2 we also consumed the inner carv1's header. I imagine that's impractical because there could be 1GiB of padding between them.

What worries me is that one could miss this line in the doc, write code that uses the roots list and happens to work with carv1 archives and is supposedly forwards compatible, but then would silently break in two weeks when it starts being fed carv2 archives.

One way to make this clearer would be to have a ReadVersion that just returns the version, and a separate ReadHeaders that can have a higher cost and always returns more information.

mvdan (approved):

v2/reader.go
masih: I wish we could fix this in root carv1 by adding extra funcs but hey.

v2/reader.go
masih: > Presumably we could always provide the roots, if for a version 2 we also consumed the inner carv1's header. I imagine that's impractical because there could be 1GiB of padding between them.

That won't be reading the `Pragma`. the intention here is to explicitly read what the spec specifies as sections in a CART v2 file. According to the spec, if version is 2 then we must never have any roots.

> What worries me is that one could miss this line in the doc, write code that uses the roots list and happens to work with carv1 archives and is supposedly forwards compatible, but then would silently break in two weeks when it starts being fed carv2 archives.

Note if version is 2, roots are nil. That's an error in CAR v1 I think, since decoding a car v1 header with empty roots would return an empty slice. I'd love to understand the workflow you can think of that makes this function error-prone. There is something here to improve I am sure of it.

> One way to make this clearer would be to have a ReadVersion that just returns the version, and a separate ReadHeaders that can have a higher cost and always returns more information.

This means that function should only take `io.ReaderAt` or would swallow up `roots` if they are present in case version is 1.

The workflow I had in mind for this is that a user can probe what version they are dealing with and proceed to read the rest of the file using `carv2.Reader` _if_ the version is 2. Otherwise they would revert back to `carv1` libraries. I am hoping folks in ignite team would help us taylor this API to be most effective. For now what's here satisfies the requirement of being able to tell what version of a CAR file one is dealing with.

mvdan: I see. Then I think this could just be "give me the version", which will always be present? :)

If one has to look at the version and do something special for carv1, I think it's fine to then have to do something extra to get the roots list, such as use the carv1 decoder.

masih: That makes sense. thanks @mvdan so you think it's OK to swallow up roots if they are present in `ReadVersion`?

mvdan: Yeah, you'd just throw them away. You could even be smart about that: if you have a ReaderAt, when you encounter the roots list in dag-cbor, I imagine you could skip the entire roots list without fully decoding all of its elements. That can be an optimization TODO for now though.

PR #92, opened 2021-06-21
Run `gofumpt -l -w .`

masih:
Format code in v2 consistently using `gofumpt -l -w .`


mvdan (approved):
As a totally unbiased person, this looks better :)

PR #100, opened 2021-06-24
Rename `ReadPragma` to `ReadVersion` and drop returning roots

masih:
Change the `ReadPragma` signature to only return the version and drop
the roots that _may_ exist if the version is 1. Because, this results in
a less error-prone API, and right now it is unclear if we want
functionality that should return roots. So, we lean towards providing
less functionality unless requested.

### DIFF
--- a/v2/reader.go
+++ b/v2/reader.go
@@ -25,7 +25,7 @@ func NewReader(r io.ReaderAt) (*Reader, error) {
 	cr := &Reader{
 		r: r,
 	}
-	if err := cr.requireV2Pragma(); err != nil {
+	if err := cr.requireVersion2(); err != nil {
 		return nil, err
 	}
 	if err := cr.readHeader(); err != nil {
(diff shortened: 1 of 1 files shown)


mvdan (approved):
Thinking outloud, we want this to take an `io.Reader` instead of a `[]byte` because for a real carv1 archive, we need to skip over the entire roots list, right?

I wonder if we could do something slightly smarter, like:

1) If the carv1 header begins with a non-empty roots list, it's a v1.
2) If the carv1 header begins with an empty roots list, then we skip it (fixed length) and read the version.

If we did that, we'd always read a fixed length, so we could just take a `[]byte`. We wouldn't support carv1-looking headers whose version is not 1, but the carv1 spec requires the version to always equal 1.

Not to do right now, just food for thought. We can move this into a v2 issue.

masih:
That's right we want to skip. 

I think we also want this to be correct for future versions maybe? in that, if roots is non-empty it doesn't necessarily mean version is 1. If we want to answer `IsVersion2` then yes we can be much cleverer, by reading the first byte alone we can rule out a lot of reading (i.e. varint(10)).

Definitely worth capturing an issue for; I'll do that. Thanks for pointing this out @mvdan

mvdan:
> I think we also want this to be correct for future versions maybe? in that, if roots is non-empty it doesn't necessarily mean version is 1.

Strictly speaking you are right; in the future there could be a carv3 that breaks that assumption. In practice, I wonder if we could make that assumption, try to keep it in the future, and save some work :) Probably not worth it, just thinking outloud.

PR #102, opened 2021-06-24
Remove dependency to CAR v1 go module

masih:
Remove dependency to CAR v1 go module, because that module depends on
later version of ipld-prime which have not been reflected in filecoin
project.

In the interest of time, we fork the code that v2 needs in v1. This
allows a more efficient implementation of some APIs that accep types
like `bufio.Reader` while helping the filecoin team push ahead with
deliverables without undergoing a big upgrade process.

On the CAR v2 side, this however means that we postpone the
implementation of Selective Car for v2 until filecoin is upgraded its
ipld-prime dependency. This is to avoid implementing an obsolete
selective car API that uses the old ipld-prime API which force user yet
another upgrade.

### DIFF
--- a/v2/blockstore/readonly.go
+++ b/v2/blockstore/readonly.go
@@ -11,10 +11,10 @@ import (
 	blocks "github.com/ipfs/go-block-format"
 	"github.com/ipfs/go-cid"
 	blockstore "github.com/ipfs/go-ipfs-blockstore"
-	carv1 "github.com/ipld/go-car"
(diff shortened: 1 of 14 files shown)


mvdan (approved):

mvdan:
We can gradually move/delete/change code as needed. None of this is exposed API anyway.

aarshkshah1992:
@mvdan @masih Is this good to go for a merge ?

PR #103, opened 2021-06-24
copy v1 license into the v2 module

mvdan:
This will allow pkgsite to render the wip-v2 module.

### DIFF
--- a/v2/LICENSE.md
+++ b/v2/LICENSE.md
@@ -1,0 +1,229 @@
+The contents of this repository are Copyright (c) corresponding authors and
+contributors, licensed under the `Permissive License Stack` meaning either of:
+
+- Apache-2.0 Software License: https://www.apache.org/licenses/LICENSE-2.0
+  ([...4tr2kfsq](https://gateway.ipfs.io/ipfs/bafkreiankqxazcae4onkp436wag2lj3ccso4nawxqkkfckd6cg4tr2kfsq))
+
+- MIT Software License: https://opensource.org/licenses/MIT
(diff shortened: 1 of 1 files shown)


v2/LICENSE.md
```diff
@@ -0,0 +1,229 @@
+The contents of this repository are Copyright (c) corresponding authors and
+contributors, licensed under the `Permissive License Stack` meaning either of:
+
+- Apache-2.0 Software License: https://www.apache.org/licenses/LICENSE-2.0
+  ([...4tr2kfsq](https://gateway.ipfs.io/ipfs/bafkreiankqxazcae4onkp436wag2lj3ccso4nawxqkkfckd6cg4tr2kfsq))
+
+- MIT Software License: https://opensource.org/licenses/MIT
+  ([...vljevcba](https://gateway.ipfs.io/ipfs/bafkreiepofszg4gfe2gzuhojmksgemsub2h4uy2gewdnr35kswvljevcba))
```
ribasushi, line 8: @lidel @aschmahmann folks are copying this license file more and more these days
Are the links legit, or should I change to something else ( dweb.link? )

mvdan: This is a straight up copy of the current v0 module in master. If the links need changing here, they need changing in lots of other places at the same time too.

masih (approved):

v2/LICENSE.md
mvdan: Merging for now. This is a WIP branch anyway, so not final in any way.

lidel: We won't be shutting gateway.ipfs.io down, but its ok to swap `gateway.ipfs.io` with `dweb.link` if  you see it.

PR #107, opened 2021-06-24
Improve documentation and fix bug in writer

masih:
Add documentation where it is sparse to help folks learn how to use the
library easier.

Unexport the ability to select codec, until we decide what codec we want
to expose. Fix all encodings of index to `IndexSorted`.

Fix a bug in writer where index is written by direct marshalling instead
of index.WriteTo which includes index codec.

Refactor in places for better readability.


mvdan (approved):
I think we might want to keep the "index algorithm" exposed in the final API, but for now SGTM.

PR #109, opened 2021-06-25
add Close methods and NewReaderMmap

aarshkshah1992:
@mvdan @masih 

- It's nice to be able to open blockstores/CARv2 Readers with just the CARv2 file paths. 
- The blockstores should have a close method that closes the backing CARv2 file.

### DIFF
--- a/v2/reader.go
+++ b/v2/reader.go
@@ -9,13 +9,32 @@ import (
 
 	"github.com/ipfs/go-cid"
 	"github.com/ipld/go-car/v2/internal/carv1"
+	"golang.org/x/exp/mmap"
 )
 
 // Reader represents a reader of CAR v2.
 type Reader struct {
-	Header Header
-	r      io.ReaderAt
-	roots  []cid.Cid
+	Header      Header
+	r           io.ReaderAt
+	roots       []cid.Cid
+	carv2Closer io.Closer
+}
+
+// NewReaderMmap is a wrapper for NewReader which opens the file at path with
+// x/exp/mmap.
+func NewReaderMmap(path string) (*Reader, error) {
+	f, err := mmap.Open(path)
+	if err != nil {
+		return nil, err
+	}
+
+	r, err := NewReader(f)
+	if err != nil {
+		return nil, err
+	}
+
+	r.carv2Closer = f
+	return r, nil
 }
 
 // NewReader constructs a new reader that reads CAR v2 from the given r.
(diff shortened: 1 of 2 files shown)


v2/reader.go
```diff
@@ -34,6 +51,10 @@ func NewReader(r io.ReaderAt) (*Reader, error) {
 	return cr, nil
 }
 
+func (r *Reader) CARv1Size() uint64 {
```
aarshkshah1992, line 54: This is nice to have for the storage market.

v2/reader.go
```diff
@@ -9,13 +9,30 @@ import (
...
+	if err != nil {
+		return nil, err
+	}
+
+	r, err := NewReader(f)
+	if err != nil {
+		return nil, err
+	}
+
+	r.closer = f
```
mvdan, line 34: I think we should remove this and use a type assertion on `r.r` being an `io.Closer` instead. That way there's less manual magic, and Close automatically does the right thing regardless of the API that was used to create the Reader.

v2/reader.go
mvdan: Perhaps there's something I'm missing, but why is `r.Carv1Size()` easier to do than `r.Header.Carv1Size`? You barely save any keystrokes, and now we have an extra bit of API to document and learn to use :) I'd rather only add more ways of accomplishing the same if they allow actually saving a significant amount of code or complexity, but this doesn't feel like it to me.

v2/reader.go
```diff
@@ -9,13 +9,30 @@ import (
...
 
 // Reader represents a reader of CAR v2.
 type Reader struct {
 	Header Header
 	r      io.ReaderAt
 	roots  []cid.Cid
+	closer io.Closer
+}
+
+func NewReaderFromCARv2File(path string) (*Reader, error) {
```
mvdan, line 23: How about `NewReaderMmap`? Every reader API takes a CARv2 file, so the current name doesn't really help the user differentiate it from NewReader. And I imagine this will always do an mmap; if we ever changed it to use os.Open instead, that would be a big change for downstreams, so it probably wouldn't be a good idea anyway.

v2/reader.go
mvdan: To clarify, I mean the type assertion should be inside the Close method.

v2/reader.go
aarshkshah1992: @mvdan Didn't realize this was exported :p My bad.

v2/reader.go
aarshkshah1992: @mvdan I agree.

mvdan:
@aarshkshah1992 want to take a look now? I've incorporated all my suggestions minus "type assert Closer".

willscott (approved):

PR #110, opened 2021-06-27
CARv2 blockstore project bug fixes, bug reports and more requirements

aarshkshah1992:
Note: This PR is only meant as a way to communicate. This is hacky and wont be merged.

- Contains bug-fixes .
- Contains bug reports -> see the tests.
- Some more requirements (see `TODOs`).

**- Also, the following simply hangs:** (@mvdan @masih )
   - Take a CARv1 file.
   - Open a CARv2 read write blockstore and write the above CARv1 to it by iterating over it.
   - Finalize the read-write blockstore .
   - Generate a read-only blocktstore on top of our shiny new CARv2 file and give it to a CARv1 Selective traversal. This simply hangs as the associated traversal/walk hangs (Note: Use the non `go-ipld-prime `CARv1 library version).
   - [This test](https://github.com/filecoin-project/go-fil-markets/blob/a8cb025698b87d9847ff1dfa9372c3caa706d4d4/storagemarket/impl/clientutils/clientutils_test.go#L61) reproduces it.
   
- Also , see https://github.com/ipld/go-car/pull/116.


v2/blockstore/readwrite_test.go
```diff
@@ -1,19 +1,202 @@
...
 	"github.com/ipld/go-car/v2/blockstore"
 
 	"github.com/ipfs/go-cid"
 	"github.com/ipld/go-car/v2/internal/carv1"
 )
 
+const unixfsChunkSize uint64 = 1 << 10
+const unixfsLinksPerLevel = 1024
+
+func TestFailure(t *testing.T) {
```
aarshkshah1992, line 32: @mvdan @masih @willscott 

These test failures are important to fix. We can go over them in our sync call.

v2/blockstore/readwrite.go
```diff
@@ -57,6 +57,9 @@ func WithIndexPadding(p uint64) Option {
 	}
 }
 
+// TODO When we are constructing a UnxFS DAG from a user file using this read-write blockstore as the staging store,
+// we might not have the root Cid will we finalise.
+// Please can we add an option to skip the root here and give it when we finalize the DAG ?
```
willscott, line 62: That requires a full re-location / re-write of the car on finalize though.

Do you know there's a single root CID and it's expected hash type? we'll probably need to do something tricky to avoid the change in length of the header that'll come from filling in the true root cid at the end.

ribasushi: It really won't matter: when you construct a dag you tart "leaves first", yet the car file for Fil-purposes must be "leaves last". Since you will be literally rewriting the position of every single block in your .car, you might as well start from he header: that is you don't *need* to know how large it is.

v2/blockstore/readonly.go
```diff
@@ -52,11 +52,14 @@ func OpenReadOnly(path string, attachIndex bool) (*ReadOnly, error) {
 
 	var idx index.Index
 	if !v2r.Header.HasIndex() {
+
 		idx, err := index.Generate(v2r.CarV1Reader())
 		if err != nil {
 			return nil, err
 		}
 		if attachIndex {
+			// TODO Shouldn't we close the v2r Reader before we open the same file here and write the Index to it.
```
aarshkshah1992, line 61: @mvdan Can this be a bug ?

v2/blockstore/readwrite.go
aarshkshah1992: @willscott @mvdan This is not a must-have. We can work our way around it by:

- Generate the UnixFS DAG using a Badger blockstore to get the root Cid.
- Generate the UnixFS DAG yet again using a CARv2 read-write blockstore now that we have the root.
- Delete the Badger blockstore.

However , this has obvious inefficiencies and we wont be able to remove _completely_ from deal making.

willscott: @ribasushi seems to be saying that data transfer is not in the canonical selector order. that would mean needing a full re-write on finalization. Can you confirm, @aarshkshah1992  if that will be true? 

If it's just that you don't know the root at the beginning, but will at the end, it's probably better for your code to have the hack to make this work: you'll start by specifying a single root of a 0 cid of the expected multi-hash type, and then we'll let you update/replace it at the time of finalization to the real root.

ribasushi: Wait... no. Grapsync *should* be doing stuff in "tarversal" order ( there is some internal buffering stuff, but that should be hidden from you... I hope )
The context was "When we are constructing a UnxFS DAG from a user file" which to me implies the local `lotus client import`. Did I misunderstand?

willscott: Ok. great. this issue of roots is in the context (i think) of the receiving side of a graphsync transfer.
I don't think we're using any of this on the client where the import happens.

willscott: (at least for this first draft)

ribasushi: > this issue of roots is in the context (i think) of the receiving side of a graphsync transfer.

Graphsync always knows the root somehow ( it needs to start validation-traversal from somewhere ). You definitely should have the root cid available by some means before you received your first ever block ( to validate that this first block is what you expect, and so on )

mvdan: We'll talk about this in the daily carv2 sync :)

mvdan:
FYI I'll be extracting bits of this into separate PRs with tests, and linking back here. I don't think we can just add the test you wrote here, because it pulls in a handful of extra IPFS dependencies, and we likely don't want those in the go.mod of go-car/v2.

aarshkshah1992:
@mvdan Oh ofcourse, sure. This PR was only meant as a way to communicate. But, can we keep it open till all the action items are resolved ?

mvdan:
Yep for sure. I'll be treating it as an issue with runnable code :)

v2/blockstore/readonly.go
```diff
@@ -165,6 +169,8 @@ func (b *ReadOnly) AllKeysChan(ctx context.Context) (<-chan cid.Cid, error) {
 
 	ch := make(chan cid.Cid, 5)
 	go func() {
+		// TODO We should close this channel to prevent a hang.
+		defer close(ch)
```
mvdan, line 173: This has been fixed in wip-v2.

v2/blockstore/readonly.go
```diff
@@ -52,11 +52,14 @@ func OpenReadOnly(path string, attachIndex bool) (*ReadOnly, error) {
...
 	var idx index.Index
 	if !v2r.Header.HasIndex() {
+
 		idx, err := index.Generate(v2r.CarV1Reader())
 		if err != nil {
 			return nil, err
 		}
 		if attachIndex {
+			// TODO Shouldn't we close the v2r Reader before we open the same file here and write the Index to it.
+			// Can you write to an already opened file and have an open reader pick it up ?
```
mvdan, line 62: This is a good question.

We can't just close the mmap reader, because we want to return it. We'd have to open it again.

Ideally we'd only "open" the file once, but I don't think that's possible: x/exp/mmap does not allow writing.

I think the best solution might be to do a regular old os.OpenFile with O_RDWR to check the header and, if the index is missing, calculate it and write it to the file. This shouldn't be slower than the mmap, given that we're reading the entire file anyway. Plus, it's a bit of an edge case; I imagine the fast path would be that the index already exists.

Then, when we're done making sure that the index exists, we close the O_RDWR file and reopen with a read-only mmap, returning the ReadOnly.

mvdan: @aarshkshah1992 thoughts? if it sounds good, I can implement the above.

To be clear, I'm not sure if writing to the file while it's open as an mmap is valid or not. But either way it seems unnecessary and a bit risky.

v2/blockstore/readwrite.go
```diff
@@ -125,6 +130,10 @@ func (b *ReadWrite) isFinalized() bool {
 	return b.header.CarV1Size != 0
 }
 
+// TODO
+// 1. Can we consider finalize to be the `Close` of the read write blockstore ?
+// 2. Are Finalize/close idempotent ? Can we call them multiple times without side-effects/errors ?
```
mvdan, line 135: 1) Finalize is not really a Close. See https://github.com/ipld/go-car/pull/114, which makes this much clearer.

2) They are not idempotent. At the current wip-v2 HEAD, calling Finalize a second time will panic, and calling ReadOnly.Close a second time will call os.File.Close a second time, which itself is documented to error on repeated calls.

Is there a particular reason you want idempotency? You could always accomplish it via something like boolean fields or https://golang.org/pkg/sync/#Once.

v2/blockstore/readwrite.go
mvdan: We can deprioritize this as per today's call.

v2/blockstore/readwrite.go
```diff
@@ -57,7 +57,12 @@ func WithIndexPadding(p uint64) Option {
 	}
 }
 
+// TODO When we are constructing a UnxFS DAG from a user file using this read-write blockstore as the staging store,
+// we might not have the root Cid will we finalise.
+// Please can we add an option to skip the root here and give it when we finalize the DAG ?
 // NewReadWrite creates a new ReadWrite at the given path with a provided set of root CIDs as the car roots.
+//
+// TODO We need an option/feature here to avoid writing/Putting duplicates.
```
mvdan, line 65: We'll make this the default behavior, to keep the set of blocks and the index in sync.

v2/car.go
```diff
@@ -137,3 +137,6 @@ func (h *Header) ReadFrom(r io.Reader) (int64, error) {
 	h.IndexOffset = binary.LittleEndian.Uint64(buf[16:])
 	return n, nil
 }
+
+// TODO: Need a utility to get a CARv2 file from a CARv1 file, basically a CARv1 -> indexed CARv2 transformer func.
```
mvdan, line 141: I'll add a high-level API for this, which would wrap an existing carv1 as a carv2 while keeping the carv1 bytes identical and building an index.

v2/blockstore/readonly.go
aarshkshah1992: @mvdan Is this done ?

mvdan: I was waiting for your reply :) If you gave the thumbs up yesterday or hours ago, I didn't notice until now - github doesn't send notification emails for emoji reactions.

I'll file an issue about this and do it, though it doesn't seem to be blocking you.

v2/car.go
mvdan: Now tracked in https://github.com/ipld/go-car/issues/119.

v2/blockstore/readwrite.go
mvdan: On 1, we've agreed to have Finalize be the read-write blockstore's Close.

On 2, we agreed that they don't need to be idempotent.

v2/blockstore/readonly.go
mvdan: Now tracked at https://github.com/ipld/go-car/issues/120.

v2/blockstore/readwrite.go
mvdan: I've started a thread on this in Slack, and will open an issue afterwards.

mvdan:
The bugs here have been solved, and all other TODOs/threads have been opened as separate issues.

v2/blockstore/readwrite.go
aarshkshah1992: Had a chat with @raulk today and re-opening this as removing Badger completely from all aspects of Deal making is an important goal of the CARv2 blockstore project which we will not be able to accomplish without this feature.

mvdan: @aarshkshah1992 please open a new issue instead, with the notes you mentioned during our daily sync. Tracking stuff to do via draft PR sub-threads is not going to work :)

mvdan:
As per slack discussion, we've once again agreed to keep "specify roots with ReadWrite.Finalize" out of scope for the first carv2 release. Closing again. Please ping me on Slack or raise an issue if that's wrong or something else comes up; I don't think we should use this PR to track stuff to do at this point.

v2/blockstore/readonly.go
masih: As per #137, more specifically https://github.com/ipld/go-car/pull/137/commits/71bc303b4ce78eea0e44a1537332933b8e3c9e0d `ReadOnly` blockstore API will no longer modify the path given. Instead, a user is encouraged to use `index.Generate` and `index.Attach` to modify an index-less CAR file, may it be v1 or v2. 

Closing #120 as a result.

PR #111, opened 2021-06-28
blockstore: close AllKeysChan channel when done

mvdan:
(see commit message)

### DIFF
--- a/v2/blockstore/readonly.go
+++ b/v2/blockstore/readonly.go
@@ -163,27 +163,28 @@ func (b *ReadOnly) AllKeysChan(ctx context.Context) (<-chan cid.Cid, error) {
 		return nil, err
 	}
 
+	// TODO: document this choice of 5, or use simpler buffering like 0 or 1.
 	ch := make(chan cid.Cid, 5)
+
 	go func() {
-		done := ctx.Done()
+		defer close(ch)
 
 		rdr := internalio.NewOffsetReader(b.backing, int64(offset))
 		for {
 			l, err := binary.ReadUvarint(rdr)
 			thisItemForNxt := rdr.Offset()
 			if err != nil {
-				return
+				return // TODO: log this error
 			}
 			c, _, err := internalio.ReadCid(b.backing, thisItemForNxt)
 			if err != nil {
-				return
+				return // TODO: log this error
 			}
 			rdr.SeekOffset(thisItemForNxt + int64(l))
 
 			select {
 			case ch <- c:
-				continue
-			case <-done:
+			case <-ctx.Done():
 				return
 			}
 		}
--- a/v2/blockstore/readwrite_test.go
+++ b/v2/blockstore/readwrite_test.go
@@ -1,10 +1,12 @@
 package blockstore_test
 
 import (
+	"context"
 	"io"
 	"math/rand"
 	"os"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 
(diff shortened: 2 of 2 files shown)


mvdan:
cc @dirkmc, I can't request a review since you're not in this repo. Maybe ask to be added to the ipld org?

v2/blockstore/readonly.go
```diff
@@ -165,7 +165,7 @@ func (b *ReadOnly) AllKeysChan(ctx context.Context) (<-chan cid.Cid, error) {
 
 	ch := make(chan cid.Cid, 5)
 	go func() {
-		done := ctx.Done()
+		defer close(ch)
```
dirkmc, line 168: I'd suggest writing out a log line when there's an error (lines 174 and 178)

v2/blockstore/readonly.go
```diff
@@ -165,7 +165,7 @@ func (b *ReadOnly) AllKeysChan(ctx context.Context) (<-chan cid.Cid, error) {
 
 	ch := make(chan cid.Cid, 5)
```
dirkmc, line 166: Where does the number 5 come from?

dirkmc (approved):
LGTM 👍 

I have a couple of non-blocking suggestions in unrelated parts of the code, but they are unrelated to this PR so may make sense to address them in a follow-up PR

v2/blockstore/readonly.go
mvdan: I wondered the same; this originates from @willscott's carbs at https://github.com/willscott/carbs/blob/1036c89e1fd753bc375d860725c9e29ab85b72ac/carbs.go#L161.

My guess is he wanted the channel to have some buffering for the sake of performance. I imagine that, unless measured, we should stick to a buffer size of 1, or no buffering. Any other buffer size should be carefully selected and documented, in my opinion :)

v2/blockstore/readonly.go
mvdan: This is a library though, so logging requires some form of API to allow being configured - and to probably be off by default. Is there a standard across PL? I know https://github.com/ipfs/go-log, but that feels too heavy to pull into relatively low-level IPLD libraries.

Maybe I just expose an option like:

```
WithLogging(func(format string, args ...interface{})) Option
```

v2/blockstore/readonly.go
ribasushi: Obligatory thread: https://twitter.com/erentantekin/status/1330957563496525824

v2/blockstore/readonly.go
dirkmc: I guess really the error should be returned to the caller somehow but I assume it's not possible to change the AllKeysChan API at this stage.
Suggest we merge this PR as a strict improvement and discuss options in an issue.

v2/blockstore/readonly.go
dirkmc: Agreed 👍

v2/blockstore/readonly.go
mvdan: This is the blockstore interface, so we indeed can't change the method signature :) I'll raise an issue about logging.

v2/blockstore/readonly.go
mvdan: I've added a TODO about this.

v2/blockstore/readonly.go
mvdan: https://github.com/ipld/go-car/issues/112

PR #113, opened 2021-06-28
blockstore: make unsupported methods panic

mvdan:
(see commit message)


mvdan:
@dirkmc @aarshkshah1992

mvdan:
The next PR will add support for using a finalized blockstore as read-only, by making Finalize return a ReadOnly.

dirkmc (approved):

PR #114, opened 2021-06-28
blockstore: make Finalize return *ReadOnly

mvdan:
(see commit message)

### DIFF
--- a/v2/blockstore/readwrite.go
+++ b/v2/blockstore/readwrite.go
@@ -120,25 +120,43 @@ func (b *ReadWrite) PutMany(blks []blocks.Block) error {
 	return nil
 }
 
-// Finalize finalizes this blockstore by writing the CAR v2 header, along with flattened index
-// for more efficient subsequent read.
-// After this call, this blockstore can no longer be used for read or write.
-func (b *ReadWrite) Finalize() error {
+// Finalize finalizes this blockstore by writing the CAR v2 header,
+// along with flattened index for more efficient subsequent read.
+//
+// After the call, this blockstore can no longer be used.
+// If a nil error is returned, the resulting blockstore can be used for
+// read-only operations.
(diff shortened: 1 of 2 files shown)


mvdan:
also @dirkmc

mvdan:
Note that ReadWrite does not have a Close method. If one just wants to write via a blockstore and close the file, it means two steps: Finalize + ReadOnly.Close. A couple of extra lines, but I think it's easy enough.

dirkmc:
I wonder if there's a way to optionally get the read-only blockstore so that the caller doesn't have to explicitly call ro.Close() after calling rw.Finalize(), if there's no error.

Would it make sense to have a separate method to get the read-only blockstore?

mvdan:
What would you call the method that doesn't return a ReadOnly blockstore?

I'd rather not call it Close, because too often people forget to check for its error when writing, e.g. via `defer bs.Close()`.

It might also be a bit confusing to have two slightly different and mutually exclusive "finalize" methods, so I thought a single one would be best. I'm open to ideas though.

dirkmc:
I see where you're coming from.

Maybe have one method called Finalize that just closes the read-write blockstore, and another called ToReadOnly that returns a read-only blockstore (that the client has to close), or something like that?

mvdan:
Actually, this PR is wrong. It returns a ReadOnly blockstore backed by an *os.File, not an mmap, so that's likely not what we want.

willscott:
> Actually, this PR is wrong. It returns a ReadOnly blockstore backed by an *os.File, not an mmap, so that's likely not what we want.

I can look at this but it sounds like i should wait until you take the next refactor

mvdan:
We've agreed that Finalize doesn't need to return a ReadOnly, so this PR is unnecessary :)

PR #115, opened 2021-06-28
consistently use CID hashes for uniqueness

mvdan:
(see commit message)

### DIFF
--- a/v2/blockstore/readonly.go
+++ b/v2/blockstore/readonly.go
@@ -2,6 +2,7 @@ package blockstore
 
 import (
 	"bufio"
+	"bytes"
 	"context"
 	"encoding/binary"
 	"fmt"
@@ -40,7 +41,6 @@ func ReadOnlyOf(backing io.ReaderAt, index index.Index) *ReadOnly {
 // If attachIndex is set to true and the index is not present in the given CAR v2 file,
 // then the generated index is written into the given path.
 func OpenReadOnly(path string, attachIndex bool) (*ReadOnly, error) {
-
 	v2r, err := carv2.NewReaderMmap(path)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


v2/index/indexsorted.go
```diff
@@ -70,16 +69,12 @@ func (s *singleWidthIndex) Unmarshal(r io.Reader) error {
...
+func (s *singleWidthIndex) Less(i int, hash []byte) bool {
+	return bytes.Compare(hash[:], s.index[i*int(s.width):((i+1)*int(s.width)-8)]) <= 0
 }
 
 func (s *singleWidthIndex) Get(c cid.Cid) (uint64, error) {
-	d, err := multihash.Decode(c.Hash())
-	if err != nil {
-		return 0, err
-	}
-	return s.get(d.Digest), nil
```
willscott, line 82: this is an on-disk change to the format of the index / needs to be reflected in spec?

I think the point here was that it'll be the same byte indicating the type of multihash on every entry, so why are we repeating that every time?

internally it seems like it should be okay to store just the digest, no?

mvdan: You're right, this breaks away from the spec.

Technically I could construct two multihashes with the same digest but different multihash codes; the blockstore compares the whole multihash so it would treat them as different, but this index only looks at the digest length and digest bytes, so it would treat them as equal.

But in practice, only storing the digest in each digest-length bucket saves space, and is going to be enough for uniqueness for basically every user. It should only cause problems if someone ends up with the same digest on different hashing algorithms, which is extremely unlikely.

I'll update the PR to revert the changes to the index.

willscott (approved):

mvdan:
I've simply reverted all changes to the index package. None of the interfaces changed there, so all tests pass.

PR #116, opened 2021-06-29
Fails to create a UnixFS DAG with a CARv2 blockstore

aarshkshah1992:
@mvdan @dirkmc

Pulls in changes from https://github.com/ipld/go-car/pull/115.

But, still see the "not found" error when I try to create a UnixFS DAG with a CARv2 read-write blockstore. The same test works fine with a in memory map-based blockstore.

See the `TestUnixFSDAGcreation `.


mvdan:
#117 fixed this.

PR #117, opened 2021-06-29
blockstore: don't error on Has not finding a CID

mvdan:
(see commit message)

### DIFF
--- a/v2/blockstore/readonly.go
+++ b/v2/blockstore/readonly.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"io"
 
@@ -84,7 +85,9 @@ func (b *ReadOnly) DeleteBlock(_ cid.Cid) error {
 // Has indicates if the store contains a block that corresponds to the given key.
 func (b *ReadOnly) Has(key cid.Cid) (bool, error) {
 	offset, err := b.idx.Get(key)
-	if err != nil {
+	if errors.Is(err, index.ErrNotFound) {
+		return false, nil
+	} else if err != nil {
 		return false, err
 	}
 	uar := internalio.NewOffsetReader(b.backing, int64(offset))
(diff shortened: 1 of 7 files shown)


mvdan:
cc @dirkmc

v2/blockstore/readonly.go
```diff
@@ -84,7 +85,9 @@ func (b *ReadOnly) DeleteBlock(_ cid.Cid) error {
 // Has indicates if the store contains a block that corresponds to the given key.
 func (b *ReadOnly) Has(key cid.Cid) (bool, error) {
 	offset, err := b.idx.Get(key)
-	if err != nil {
+	if errors.Is(err, index.ErrNotFound) {
+		return false, nil
+	} else if err != nil {
 		return false, err
 	}
```
aarshkshah1992, line 92: So, what you are saying is that if the Index dosen't have a cid, the blockstore dosen't have the cid but there's no error.

The other code path is that "something went wrong" when trying to lookup the Index.

aarshkshah1992: Why was this resulting into "not found" when building the UnixFS DAG ?

mvdan: This is the error coming from the index. The index only has a "get offset" method, which returns a non-nil error if a CID is not in the index, or if something went truly wrong - such as a corrupted index. In the former case, where everything went fine but a CID is simply missing, we get an ErrNotFound. In this non-fatal case, the blockstore interface wants Has to return a nil error, and to simply report that with a "false" return value.

> Why was this resulting into "not found" when building the UnixFS DAG ?

I added some debug prints to the blockstore. It seems like ipld-prime's "async commit" to the blockstore did a bunch of bs.Has calls before bs.Put, perhaps as a means to avoid duplicating work. Those were returning a seemingly fatal error of "not found", so the whole thing failed immediately.

aarshkshah1992 (approved):
Thanks @mvdan .

PR #122, opened 2021-06-29
blockstore: make it safe for concurrent use

mvdan:
(see commit message)

### DIFF
--- a/v2/blockstore/readonly.go
+++ b/v2/blockstore/readonly.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"sync"
 
 	blocks "github.com/ipfs/go-block-format"
 	"github.com/ipfs/go-cid"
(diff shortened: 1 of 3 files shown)


v2/blockstore/readonly.go
```diff
@@ -152,6 +170,9 @@ func (b *ReadOnly) PutMany([]blocks.Block) error {
 
 // AllKeysChan returns the list of keys in the CAR.
 func (b *ReadOnly) AllKeysChan(ctx context.Context) (<-chan cid.Cid, error) {
+	b.mu.RLock()
+	defer b.mu.RUnlock()
+
```
dirkmc, line 175: Do we also need to take the read lock inside the go-routine?

dirkmc (commented):
Do we need to take a lock in the Roots and Close methods?

v2/blockstore/readonly.go
mvdan: Yikes, good catch. We probably want to grab the mutex here, and release it when the other goroutine ends.

mvdan: Done.

aarshkshah1992:
LGTM. Just need to address @dirkmc 's comment here.

mvdan:
It's already addressed, so merging.

dirkmc:
@mvdan do we need to take a lock in the Roots and Close methods?

mvdan:
Neither is part of the blockstore interface, so my intuition is that they don't need locking. Close should never be called concurrently to reads, as far as I can tell. And the roots are set at the beginning and never modified, so that should never be racy.

Really, the only racy bits here are the index and the io.Writer for appending blocks to the carv1.

PR #125, opened 2021-06-30
Tests to verify CARv2 blockstore dosen't put duplicates and has the required de-duped blocks

aarshkshah1992:
- Generates a UnixFS DAG containing duplicate blocks -> payload has been written to ensure this happens.
- Generates a CARv2 file from the UnixFS DAG using the CARv2 read-write blockstore.
- Iterates over the CARv1 payload of the CARv2 file to ensure there are no duplicates in it **_(this step currently fails)_**.
- Does a CARv1 selective CAR traversal on the UnixFS DAG using an inmemory blockstore -> verified that this produces no duplicates.
- Assert that the de-duped blocks seen in the above step are the same as when we iterate the CARv1 payload of the CARv2 file.


mvdan:
Fixed by https://github.com/ipld/go-car/pull/127.

PR #126, opened 2021-06-30
add WrapV1File

mvdan:
(see commit message)


mvdan:
cc @dirkmc @aarshkshah1992

mvdan:
Ah, I forgot to also have the example verify that the original CARv1 and new inner CARv1 are exactly equal. I'll add that in a minute.

v2/writer.go
```diff
@@ -130,3 +131,65 @@ func (w *Writer) writeIndex(writer io.Writer, carV1 []byte) (int64, error) {
...
+
+	// TODO: also expose WrapV1(io.ReadSeeker, io.Writer),
+	// once index.Generate takes a ReadSeeker.
+
+	// We don't use mmap.Open, so we can later use io.Copy.
+	f1, err := os.Open(srcPath)
+	if err != nil {
+		return err
+	}
+	defer f1.Close()
```
dirkmc, line 151: Should we check the err return value of f1.Close()?

v2/writer.go
```diff
@@ -130,3 +131,65 @@ func (w *Writer) writeIndex(writer io.Writer, carV1 []byte) (int64, error) {
...
+		return err
+	}
+	if err := index.WriteTo(idx, f2); err != nil {
+		return err
+	}
+
+	if err := f2.Close(); err != nil {
+		return err
+	}
+	return nil
```
dirkmc, line 194:
```suggestion
	return f2.Close()
```

v2/writer.go
```diff
@@ -130,3 +131,65 @@ func (w *Writer) writeIndex(writer io.Writer, carV1 []byte) (int64, error) {
...
+		return err
+	}
+
+	// Only create the destination CARv2 when we've gathered all the
+	// information we need, such as the index and the CARv1 size.
+	f2, err := os.Create(dstPath)
+	if err != nil {
+		return err
+	}
+	defer f2.Close()
```
dirkmc, line 173: It seems like f2.Close() is called twice (see line 191)

v2/writer.go
mvdan: I don't think so, given that we only open it for reading.

v2/writer.go
mvdan: Your version is shorter, but loses a bit of consistency: it's helpful to always have the error returns in separate branches, so that the end of the body is always the "all went well" case.

For example, if I wanted to add extra code at the end in the future, with my version the diff will just be added lines, but with yours you'll also have to refactor the last error return.

All that said, I don't think this is a big deal, so if you feel strongly about it I can change it :)

v2/writer.go
mvdan: Yep, this is on purpose - duplicate Close calls might be no-ops or might error, but we don't care as we discard the error here anyway. We want to check the close error if everything else went fine, but we also don't want to leak an unclosed f2 if something went wrong, such as the io.Copy error return.

So this defer saves us from the latter - closing the file if we don't get to the end of the func. And since we do that when something went wrong... we don't care about this close error.

mvdan:
> Ah, I forgot to also have the example verify that the original CARv1 and new inner CARv1 are exactly equal. I'll add that in a minute.

This is now done. I've also switched the example to use Reader instead of the blockstore, so that we can actually check the inner CARv1's bytes.

v2/writer.go
dirkmc: Makes sense - may be worth adding also a comment with this explanation

dirkmc (approved):

v2/writer.go
mvdan: Sure, will do. I'm kinda used to this pattern at this point, personally, but I agree that a comment can't hurt. I'm not sure what other people do for files they write to, unless they ignore close errors or potentially leak open files :)

dirkmc: I use a named return value for the error, but I prefer the method you've used, named return values are bug-prone.

mvdan: Named result parameters can be OK for docs, but big +1 to avoiding "naked returns" and setting results in magic ways.

mvdan:
Index marshaling is now deterministic. @dirkmc care to re-review that bit?

Also FYI @willscott

dirkmc (approved):
LGTM 👍

PR #127, opened 2021-07-01
blockstore: add option to deduplicate by CID

mvdan:
(see commit message)

### DIFF
--- a/v2/blockstore/readwrite.go
+++ b/v2/blockstore/readwrite.go
@@ -33,6 +33,8 @@ type ReadWrite struct {
 	ReadOnly
 	idx    *index.InsertionIndex
 	header carv2.Header
+
+	dedupCids bool
 }
 
 // TODO consider exposing interfaces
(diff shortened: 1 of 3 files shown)


mvdan:
FYI @aarshkshah1992 I verified that https://github.com/ipld/go-car/pull/125 passes with this option. Thanks again for providing reproducible tests.

v2/blockstore/readwrite.go
```diff
@@ -52,6 +54,15 @@ func WithIndexPadding(p uint64) Option {
 	}
 }
 
+// WithCidDeduplication makes Put calls ignore blocks if the blockstore already
+// has the exact same CID.
+// This can help avoid redundancy in a CARv1's list of CID-Block pairs.
+//
+// Note that this compares whole CIDs, not just multihashes.
+func WithCidDeduplication(b *ReadWrite) {
```
aarshkshah1992, line 62: @mvdan This is not the default ?

mvdan: It's still unclear what the default semantics of this blockstore will be. [This thread](https://protocollabs.slack.com/archives/C01MFLTLKTP/p1625142319426200?thread_ts=1624963667.404300&cid=C01MFLTLKTP) goes into some detail.

For now, to unblock us all, I'm keeping this opt-in, and keeping the default behavior relatively close to how go-ipfs implements its blockstores.

mvdan: I've raised https://github.com/ipld/go-car/issues/130 as a reminder before the release.

dirkmc (approved):
LGTM 👍

PR #134, opened 2021-07-05
avoid exposing io.SectionReader in the API

mvdan:
(see commit message)

### DIFF
--- a/v2/reader.go
+++ b/v2/reader.go
@@ -84,15 +84,22 @@ func (r *Reader) readHeader() (err error) {
 	return
 }
 
+// SectionReader implements both io.ReadSeeker and io.ReaderAt.
+// It is the interface version of io.SectionReader, but note that the
+// implementation is not guaranteed to be an io.SectionReader.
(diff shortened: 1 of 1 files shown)


v2/reader.go
```diff
@@ -84,15 +84,22 @@ func (r *Reader) readHeader() (err error) {
 	return
 }
 
+// SectionReader implements both io.ReadSeeker and io.ReaderAt.
+// It is the interface version of io.SectionReader, but note that the
+// implementation is not guaranteed to be an io.SectionReader.
+type SectionReader interface {
+	io.Reader
+	io.Seeker
+	io.ReaderAt
```
masih, line 93: Should we add `io.ReadSeeker` here while at it?

masih (approved):

v2/reader.go
mvdan: :shrug: I thought about it but I wasn't sure which way was nicer. I thought listing the three single-method interfaces might be a bit easier for the code reader

PR #136, opened 2021-07-05
add carv2.WrapV1 based on io interfaces

mvdan:
(see commit message)


mvdan:
FYI @raulk @willscott I ended up not touching any of the APIs taking io.ReaderAt, except for index.Generate. Its implementation gets a bit easier, and allows for WrapV1 to exist without hackery (such as implementing an io.ReaderAt on top of a io.ReadSeeker...).

mvdan:
Also FYI @masih this removes a couple of uses of OffsetReader, which I think is a good thing - the fewer io layers we have, the better.

mvdan:
Also, the change of API in index.Generate just works with the rest of the codebase, because CARv1Reader already implements ReadSeeker, too :)

masih (approved):
Nice 🍻

PR #137, opened 2021-07-05
Rename `ReadOnlyOf` to `NewReadOnly` and accept both v1 and v2 payload

### LINKED ISSUE #128: blockstore: add an API to open either v1 or v2 as a read-only blockstore

masih:
Rename the `blockstore.ReadOnlyOf` to `blockstore.NewReadOnly` for
better readability and consistent naming.

Allow the function to accept both v1 and v2 payloads and "do the right
thing" depending on the input.

Optionally, allow the caller to supply an index that, if present, will
overrider any existing index. Otherwise an in-memory index is generated.

Note, `blockstore.OpenReadOnly` only accepts v2 payload. This PRs also
changes this function to also accept both versions for consistency, and
drop modification of given path, delegating any writing to the
write-specific APIs like `index.Attach` and `index.Generate`.

This PR depends on #136 for changes introduced to `index.Generate` taking `io.ReadSeeker`. 
See d62e882 commit message.

Fixes #128
Depends on #136


mvdan:
SGTM. I imagine you could replace much of OpenReadOnly's code with NewReadOnly, now.

masih:
> SGTM. I imagine you could replace much of OpenReadOnly's code with NewReadOnly, now.

That's right; was planning to do this in a separate PR but since it's mostly removing code I'll do that here.

mvdan (approved):

PR #138, opened 2021-07-05
Replace `binary.Put/ReadUvarint` with `go-varint`

### LINKED ISSUE #135: ensure we replace binary.Put/ReadUvarint with go-varint before the v2 release
We might want to replace all instances of binary.PutUvarint and binary.ReadUvarint in carv2 with https://github.com/multiformats/go-varint before the release; otherwise we are not guaranteed to have minimal encoding of varints, which is useful for determinism

Most of our uses are Reads, not Puts, but even then - go-varint will error if the incoming varint uses more bytes than it should, which is a good thing. We have a few Puts in e.g. index encoding.

masih:
To guarantee minimal encoding.

Fixes #135

### DIFF
--- a/v2/index/index.go
+++ b/v2/index/index.go
@@ -2,10 +2,11 @@ package index
 
 import (
 	"bufio"
-	"encoding/binary"
 	"fmt"
 	"io"
 	"os"
+
+	"github.com/multiformats/go-varint"
 
 	internalio "github.com/ipld/go-car/v2/internal/io"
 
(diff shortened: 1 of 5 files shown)


v2/index/index.go
```diff
@@ -82,9 +83,8 @@ func Attach(path string, idx Index, offset uint64) error {
 // The written bytes include the index encoding.
 // This can then be read back using index.ReadFrom
 func WriteTo(idx Index, w io.Writer) error {
-	buf := make([]byte, binary.MaxVarintLen64)
-	b := binary.PutUvarint(buf, uint64(idx.Codec()))
-	if _, err := w.Write(buf[:b]); err != nil {
+	b := varint.ToUvarint(uint64(idx.Codec()))
```
mvdan, line 86: FYI go-uvarint still has PutUvarint - I think it's slightly better if we're in charge of the []byte allocation, because it's easier for the compiler to put that on our stack and avoid an allocation

v2/internal/carv1/util/util.go
```diff
@@ -55,17 +54,16 @@ func LdSize(d ...[]byte) uint64 {
 	for _, s := range d {
 		sum += uint64(len(s))
 	}
-	buf := make([]byte, 8)
-	n := binary.PutUvarint(buf, sum)
-	return sum + uint64(n)
+	s := varint.UvarintSize(sum)
```
mvdan, line 57: nice.

mvdan (approved):

Issue #140, opened 2021-07-05
add support for null-padded carv1 payloads


mvdan:
Filecoin has been using a fork of the carv1 library with commit 11b6074b6d4d6b175e42d24003c1ecacbc48917a, which allows decoding carv1 files that end with null padding. They use this because their crypto proofs require power-of-two payload lengths.

Spec-wise, this seems valid. A CARv1 is a series of "frames", in the form of:

* cid+block length as a varint
* cid
* block

So, if such a frame begins with the zero byte, then we should simply skip to the next byte, because we have a varint of value 0 for the cid+block length, and thus we have nothing else to do. It's reasonable to simply ignore the frame entirely, because there's no CID and no block data.

There's only one place in the carv2 library where we decode CARv1 frames: `index.Generate`. Right now this padding makes it error, because it unconditionally reads a CID, even if the frame size is zero. In that way, it incorrectly implements the spec; it shouldn't attempt to read CID bytes when it knows a frame is of zero length. So we definitely have something to fix in carv2.

The CARv1 spec doesn't explicitly allow this kind of empty CID plus empty block, but it's not explicitly disallowed either. It does contain a section on padding, however:

> The CAR format contains no specified means of padding to achieve specific total archive sizes or internal byte offset alignment of block data. Because it is not a requirement that each block be part of a coherent DAG under one of the roots of the CAR, dummy block entries may be used to achieve padding. Such padding should also account for the size of the length-prefix varint and the CID for a section. All sections must be valid and dummy entries should still decode to valid IPLD blocks.

I think the ship has sailed in terms of forbidding this kind of padding in the CARv1 spec, given that Filecoin has been using this for some time. We could tell them to "migrate" all of those null-padded CARv1 files, so that they also store them alongside a "padding offset" and cut the payload when using the carv2 library, effectively producing an EOF error when the library reaches the padding.

However, this would come at a relatively high cost to them (migrations are work), whereas on our side it's fairly trivial to amend one line to handle zero-length frames in a better way. What I propose is to skip the zero-length frames, effectively meaning that `index.Generate` would skip the entire null padding one byte at a time.

That would be pretty slow, but it would work. Later on, we have two options:

A) In the CARv1 spec, specify that a decoder is allowed to treat a frame beginning with a null byte as an EOF, discarding the rest of the input. This would allow us to quickly exit and stop reading bytes.

[...]

PR #141, opened 2021-07-05
index: add support for null paddings in Generate

mvdan:
(see commit message)

### DIFF
--- a/v2/index/generator.go
+++ b/v2/index/generator.go
@@ -60,6 +60,13 @@ func Generate(v1 io.ReadSeeker) (Index, error) {
 			return nil, err
 		}
 
+		// Null padding; treat zero-length frames as an EOF.
+		// They don't contain a CID nor block, so they're not useful.
+		// TODO: Amend the CARv1 spec to explicitly allow this.
+		if length == 0 {
+			break
+		}
+
 		// Grab the CID.
 		n, c, err := cid.CidFromReader(v1)
 		if err != nil {
--- a/v2/blockstore/readonly.go
+++ b/v2/blockstore/readonly.go
@@ -239,16 +239,22 @@ func (b *ReadOnly) AllKeysChan(ctx context.Context) (<-chan cid.Cid, error) {
 
 		rdr := internalio.NewOffsetReadSeeker(b.backing, int64(offset))
 		for {
-			l, err := varint.ReadUvarint(rdr)
+			length, err := varint.ReadUvarint(rdr)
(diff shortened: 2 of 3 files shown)


masih (approved):

mvdan:
Actually, we need AllKeysChan to explicitly support it too. I'll update the PR.

v2/index/generator.go
```diff
@@ -60,6 +60,17 @@ func Generate(v1 io.ReadSeeker) (Index, error) {
...
 
+		// Ignore frames whose length is zero.
+		// They don't contain a CID nor block, so they're not useful.
+		// Plus, some existing CARv1 users append null paddings to files,
+		// so it's helpful if we skip them here too.
+		// TODO: In the future, the CARv1 spec should allow us to skip
+		// the rest of the input here. If not, we should do something
+		// more performant here, such as reading zeros in chunks.
+		if length == 0 {
+			continue
```
ribasushi, line 71: As mentioned in our meeting: I would *vastly* prefer this to be a `break`

Unless I am misunderstanding the codeflow...

mvdan: See the TODO :) That's pending an addition to the CARv1 spec to explicitly say that treating a null byte as an EOF is okay.

If we all agree that will definitely go into the carv1 spec, then I can make the change here, but I'm hesitant to make v2.0.0 behave in a way that's not following the spec.

mvdan: Then again, it seems like we all agree the spec should be changed that way, so I'll go ahead and change it for a break.

@rvagg please say if you disagree

rvagg: As per https://github.com/ipld/go-car/issues/140#issuecomment-874437177 I think we could make it a CARv1 reader thing, but as an opt-in when opening a CAR—signal that a `0==EOF`, otherwise `0==err`. It's not great to have Filecoin continuing to use a fork and if length data has been discarded then there's not much we can do (I did raise this very early on and iirc I was told that length data persisted through the deal!).

aarshkshah1992:
@masih We've tested this and it works. Please can we merge this ?

mvdan:
@masih can you reply to Rod's comments first? I don't have the Filecoin context.

v2/index/generator.go
masih: Agreed. I think we need to flex the spec to not break v1 considering existing usage pattern; there is no getting away from that.

masih (approved):

Issue #143, opened 2021-07-08
Client should be able to get a detached Index from a CARv1 or CARv2 file


aarshkshah1992:
Currently, if a client wants to to generate/get just the Index for a given CAR file, the client needs to write a lot of boilerplate, especially if the client dosen't know whether what it has is a CARv1 or an indexed CARv2. We don't need this feature for UnIndexed CARv2s and leave it for you guys to decide if you want to support that too for now.

```
Assume reader here is an `io.ReadSeeker` and we don't know whether the stream is a CARv1 or a CARv2:

	// determine CAR version and Index accordingly.
	ver, err := car.ReadVersion(reader)
	if err != nil {
		return
	}
	// seek to start of file as reading the version above will change the offset.
	if _, err := reader.Seek(0, io.SeekStart); err != nil {
		return
	}
	
	var idx carindex.Index
	switch ver {
	case 2:
		carreader, err := car.NewReader(reader)
		if err != nil {
			return
		}
		if has := carreader.Header.HasIndex(); !has {
			return
		}
		idx, err = carindex.ReadFrom(carreader.IndexReader())
		if err !=  nil {
			return
		}
	case 1:
		cidx,err  := carindex.Generate(reader)
		if err != nil {
			return
		}
		idx = cidx
```

It would be great to have an API that takes either a CARv1/v2 stream and can generate/detach the Index from it and return it.

masih:
Fixed in #144 by adding `ReadOrGenerateIndex(rs io.ReadSeeker) (index.Index, error)` that accepts both v1 and v2 payload,  generating index if it is not present.

PR #144, opened 2021-07-08
Implement read or generate index from CAR accepting v1 or v2

masih:
A user has to undergo a lot of boilerplate code to generate or get an
index where there is a CAR file, the version for which is unknown.
Provide a function that accepts either v1 or v2 payload and reads the
index if present, otherwise generates it for the user. The generated
index lives entirely in memory and will not depend on the given
`io.ReadSeeker` to lookup indices.

Implement tests that check index is returned for v1 or v2 files. Add a
sample file with an imaginary version of 42 for testing purposes to
assert that the indexing function will error if the given file is not v1
or v2.

Fixes https://github.com/ipld/go-car/issues/143

### DIFF
--- a/v2/reader.go
+++ b/v2/reader.go
@@ -129,12 +129,12 @@ var _ io.ReaderAt = (*readSeekerAt)(nil)
 
 type readSeekerAt struct {
 	rs io.ReadSeeker
-	mu sync.RWMutex
+	mu sync.Mutex
 }
 
 func (rsa *readSeekerAt) ReadAt(p []byte, off int64) (n int, err error) {
-	rsa.mu.RLock()
-	defer rsa.mu.RUnlock()
+	rsa.mu.Lock()
+	defer rsa.mu.Unlock()
 	if _, err := rsa.rs.Seek(off, io.SeekStart); err != nil {
 		return 0, err
 	}
--- a/v2/reader.go
+++ b/v2/reader.go
@@ -4,6 +4,9 @@ import (
 	"bufio"
 	"fmt"
 	"io"
+	"sync"
+
+	"github.com/ipld/go-car/v2/index"
 
 	internalio "github.com/ipld/go-car/v2/internal/io"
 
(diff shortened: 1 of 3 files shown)


v2/reader.go
```diff
@@ -121,3 +124,62 @@ func ReadVersion(r io.Reader) (version uint64, err error) {
 	}
 	return header.Version, nil
 }
+
+var _ io.ReaderAt = (*readSeekerAt)(nil)
+
+type readSeekerAt struct {
+	rs io.ReadSeeker
+	mu sync.RWMutex
```
aarshkshah1992, line 132: Shouldn't this be a Mutex instead of an RWMutex ?

masih: doh indeed 🍻

aarshkshah1992 (approved):
lgtm.

v2/reader.go
```diff
@@ -121,3 +124,62 @@ func ReadVersion(r io.Reader) (version uint64, err error) {
 	}
 	return header.Version, nil
 }
+
+var _ io.ReaderAt = (*readSeekerAt)(nil)
+
+type readSeekerAt struct {
```
willscott, line 130: should this live in `internal/io`?

v2/reader.go
```diff
@@ -121,3 +124,62 @@ func ReadVersion(r io.Reader) (version uint64, err error) {
...
+}
+
+// ReadOrGenerateIndex accepts both CAR v1 and v2 format, and reads or generates an index for it.
+// When the given reader is in CAR v1 format an index is always generated.
+// For a payload in CAR v2 format, an index is only generated if Header.HasIndex returns false.
+// An error is returned for all other formats, i.e. versions other than 1 or 2.
+//
+// Note, the returned index lives entirely in memory and will not depend on the
+// given reader to fulfill index lookup.
+func ReadOrGenerateIndex(rs io.ReadSeeker) (index.Index, error) {
```
willscott, line 151: there should be a way to do this on just an `io.Reader` without ever needing to seek, but we would end up needing to use internal functions to jump to the internal (post-header) portion of index generation when we encounter a car v1 header.

v2/reader.go
masih: This reader is not reused elsewhere. Happy to move it there if you think it might get reused elsewhere.

willscott: we have 3 different io interfaces we change between: `Reader`, `ReaderAt`, and `ReadSeeker`. having the internal conversions between them all in one place seems reasonable

v2/reader.go
masih: I agree; this should be doable with an `io.Reader` alone. `Seek` is mainly used for skipping over blocks when generating index, otherwise we would have to read and discard while iterating through blocks during indexing. 

Would it make sense to capture an issue and mark it as a P3 deferring implementation unless folks explicitly ask for it?

v2/reader.go
masih: https://github.com/ipld/go-car/issues/145

v2/reader.go
willscott: yeah. that works. I'll leave relative priority to you. I suspect in practice it doesn't matter _too_ much, but i can imagine situations where you're streaming a car onto disk, and want to generate an index while you're at it. Being able to pipe through this with just an `io.Reader` is likely the way you'd hope to do that, so we should eventually relax the interface here to what we actually need.

masih: https://github.com/ipld/go-car/issues/146

Issue #118, opened 2021-06-29
v2: don't expose bits of API that shouldn't be used directly


mvdan:
For example, the insertion index is only used by the incremental read-write blockstore, and is not specced, so it should never be used while encoding CARv2 files. Having it as part of the public API is dangerous - if someone misses that detail, they could encode a CARv2 with such an index directly, which would be unspecified and fragile.

masih:
Done

PR #147, opened 2021-07-08
Implement ReadWrite blockstore resumption

### LINKED ISSUE #98: Read/Write blockstore should support resuming
When writing to a blockstore, the write might fail before the blockstore writes are "finalised/flushed". The API should support resuming writes, picking up where it left off. 

This is a _must_ for Ignite requirements.

masih:
* Implement an option for read-write blockstore, that if enabled the
write can resume from where the writer left off. For resumption to work
the `WithResumption` option needs to be set explicitly. Otherwise, if
path to an existing file is passed, the blockstore construction will
return an error. The resumption requires the roots passed to constructor
as well as padding options to be identical with roots in file.
Resumption only works on paths where at least V2 pragma and CAR v1
header was successfully written onto the file. Otherwise an error is
returned.

* Implement resumption test that verifies files resumed from match
expected  header, data and index.

* Implement a CAR v1 equals function to check if two given headers are
identical. This implementation requires exact ordering of root elements.
A TODO is left to relax the exact ordering requirement.

* Implement Seeker in internal offset writer in order to forward offset
of CAR v1 writer within a resumed read-write blockstore after
resumption. The offset of the writer needs to be set to the latest
written frame in order for consecutive writes to be at the right offset.

* Fix bug in offset read seeker in internal IO, where seek and returned
position was not normalized by the base. Reflect the fix in read-only
blockstore AllKeysChan where reader was twisted to work. We now read the
 header to get its size, then seek past it to then iterate over blocks
 to populate the channel.

* Add TODOs in places to make treating zero-length frames as EOF
optional; See #140 for context.

* Run `gofumpt -l -w .` on everything to maintain consistent formatting.

Fixes #98


v2/blockstore/readwrite.go
```diff
@@ -59,50 +66,164 @@ func WithIndexPadding(p uint64) Option {
 // This can help avoid redundancy in a CARv1's list of CID-Block pairs.
 //
 // Note that this compares whole CIDs, not just multihashes.
-func WithCidDeduplication(b *ReadWrite) {
+func WithCidDeduplication(b *ReadWrite) { // TODO should this take a bool and return an option to allow disabling dedupliation?
```
willscott, line 69: the current is setup (default off, option for on) seems good enough for the first pass

v2/blockstore/readwrite.go
```diff
@@ -59,50 +66,164 @@ func WithIndexPadding(p uint64) Option {
...
+// This option is set to false by default, i.e. disabled. When disabled, the file at path must not
+// exist, otherwise an error is returned upon ReadWrite blockstore construction.
+//
+// When this option is set to true the existing data frames in file are re-indexed, allowing the
+// caller to continue putting any remaining blocks without having to re-ingest blocks for which
+// previous ReadWrite.Put returned successfully.
+//
+// Resumption is only allowed on files that satisfy the following criteria:
+//   1. start with a complete CAR v2 car.Pragma.
+//   2. contain a complete CAR v1 data header with identical roots, starting at offset specified by
```
willscott, line 85: clarify what we mean by `identical roots`

v2/blockstore/readwrite.go
```diff
@@ -59,50 +66,164 @@ func WithIndexPadding(p uint64) Option {
...
-	if _, err := f.WriteAt(carv2.Pragma, 0); err != nil {
-		return nil, err
-	}
 
-	v1Header := &carv1.CarHeader{
-		Roots:   roots,
-		Version: 1,
+	fFlag := os.O_RDWR | os.O_CREATE
+	if !b.resume {
+		fFlag = fFlag | os.O_EXCL
```
willscott, line 118: don't we want to be exclusive in the resume case as well?

v2/blockstore/readwrite.go
```diff
@@ -59,50 +66,164 @@ func WithIndexPadding(p uint64) Option {
...
+	offset, err := carv1.HeaderSize(header)
+	if err != nil {
+		return err
+	}
+	frameOffset := int64(0)
+	if frameOffset, err = v1r.Seek(int64(offset), io.SeekStart); err != nil {
+		return err
+	}
+
+	for {
```
willscott, line 192: i would at least pull this utility into the index package rather than having it inline here

v2/internal/carv1/car.go
```diff
@@ -209,3 +210,8 @@ func loadCarSlow(s Store, cr *CarReader) (*CarHeader, error) {
 		}
 	}
 }
+
+func (h CarHeader) Equals(other CarHeader) bool {
+	// TODO should headers with same roots in different order be equal?
```
willscott, line 215: i'd probably lean towards 'yes' - or at the least i think we have a canonical order they should get written into and in practice it won't matter since there are very very few cases where there isn't exactly one root

v2/blockstore/readwrite.go
masih: In the context of finalising the interfaces, do you think we should keep this function as is, instead of taking a `bool` for the v2 release? If released we need to keep it for the lifetime of v2 and introduce a `WithCidDeduplicationDisabled` or something like that if we want to allow disabling of deduplication.

On a side note, other option functions in this file return `Option` so whichever we decide it would be good to keep them consistent.

v2/blockstore/readwrite.go
masih: No, because I explicitly want `ReadWrite` constructor to return an error, when resumption is set to false but the file already exists. I have documented this on resumption option [here](https://github.com/ipld/go-car/blob/debbc7edddb6fdf161663d7e6f7311df9fe763f9/v2/blockstore/readwrite.go#L76-L77). 

The rationale is to avoid overriding existing files by accident.

v2/blockstore/readwrite.go
masih: That makes sense. I thought I do this as part of #118 where most of the index types need to be unexposed and moved back to internal packages. 

Doing this now would either mean exposing more things in `index` package that we need to unexpose later or making this PR much larger by also doing #118.

If you agree, I'll add a TODO here and attend to it in a separate PR.

v2/blockstore/readwrite.go
masih: Following the same rationale, altered the code slight to avoid creating files when resumption is enabled in https://github.com/ipld/go-car/pull/147/commits/e12754dd76e8d076f6302c13ed0127afd2de63b1.

masih:
@aarshkshah1992 @dirkmc  Thank you for the feedback on this; here is a summary:
* By default resume if file exists, removing the need to explicitly set the resumption flag to true for it to work.
* On `Finalize` write the CAR v2 header last. This reduces the chances of having unresumable files, since header size is much smaller than index.
* Improve docs by explicitly stating on which API calls the resumption conditions are met.

v2/blockstore/readwrite.go
willscott: being consistent with an `Option` seems good. i don't think i have strong feelings about whether there's a boolean argument

willscott (approved):

masih:
> @aarshkshah1992 @dirkmc Thank you for the feedback on this; here is a summary:
> 
> * By default resume if file exists, removing the need to explicitly set the resumption flag to true for it to work.
> * On `Finalize` write the CAR v2 header last. This reduces the chances of having unresumable files, since header size is much smaller than index.
> * Improve docs by explicitly stating on which API calls the resumption conditions are met.

I removed the resumption option completely now that it is attempted by default. 
I thought we will add the option to disable it when needed, since it seems redundant just now. 

I have also made the error when resuming on finalised files explicit. 

Merging; thank you all for the reviews.

v2/blockstore/readwrite.go
```diff
@@ -59,48 +65,187 @@ func WithIndexPadding(p uint64) Option {
...
+	// Try and resume by default if the file exists.
+	resume := true
+	if _, err := os.Stat(path); err != nil { // TODO should we use stats to avoid resuming from files with zero size?
+		if os.IsNotExist(err) {
+			resume = false
+		} else {
+			return nil, err
+		}
+	}
+	f, err := os.OpenFile(path, os.O_RDWR|os.O_CREATE, 0o666) // TODO: Should the user be able to configure FileMode permissions?
```
masih, line 110: @masih Open once then stat on the open file, because it is possible for the file to have moved in between stat and open calls.

v2/blockstore/readwrite.go
```diff
@@ -59,48 +65,187 @@ func WithIndexPadding(p uint64) Option {
...
+		b.idx.InsertNoReplace(c, uint64(frameOffset))
+
+		// Seek to the next frame by skipping the block.
+		// The frame length includes the CID, so subtract it.
+		if frameOffset, err = v1r.Seek(int64(length)-int64(n), io.SeekCurrent); err != nil {
+			return err
+		}
+	}
+	// Seek to the end of last skipped block where the writer should resume writing.
+	_, err = b.carV1Writer.Seek(frameOffset, io.SeekStart)
```
masih, line 247: @masih resumption should be used with deduplicated puts, so puts that got resumed don't get written again.

We can add this to the doc and or warn people about it unless they explicitly want duplicate blocks in there.

v2/internal/carv1/car.go
```diff
@@ -209,3 +209,39 @@ func loadCarSlow(s Store, cr *CarReader) (*CarHeader, error) {
 		}
 	}
 }
+
+// Equals checks whether two headers are equal.
+// Two headers are considered equal if:
+//   1. They have the same version number, and
+//   2. They contain the same root CIDs in any order.
+func (h CarHeader) Equals(other CarHeader) bool {
```
masih, line 217: @masih Rename `Equal` to something else since `Equal` has a specific meaning in go, more in the lines of `reflect.DeepEqual` where order of elements matter.

PR #148, opened 2021-07-09
Test to reproduce block not found error

aarshkshah1992:
See `TestCarIndexBug `. Comments inline to explain the test.

We've seen similar bugs before which were solved with #115 and #117. Maybe there's a hint there ?


masih:
Closing, since the `not found` error is expected. 

The test creates a CARv2 payload using `ReadWrite` blockstore, re-orders the blocks in its payload using CAR `SelectiveCar` to produce a deterministic CARv1, _but_ uses the index originally generated by the `ReadWrite` blockstore to look blocks up, where offsets in the index don't match up with the (now re-ordered) CARv1 blocks. This indeed would result in `not found` error.

For reference, the tests will pass if the original CAR v1 payload is passed instead of the deterministically ordered payload:

```diff
Index: v2/blockstore/broken_test.go
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- v2/blockstore/broken_test.go	(revision 824ee3b7e56bf45d12a22c128a1769298c9c4ad3)
+++ v2/blockstore/broken_test.go	(date 1625820779976)
@@ -62,7 +62,14 @@
 	require.NoError(t, err)
 	defer os.Remove(f2.Name())
 	defer f2.Close()
-	_, err = io.Copy(f2, carBuf)
+
+	open, err := os.Open(carV2FilePath)
+	require.NoError(t, err)
+	reader, err := carv2.NewReader(open)
+	require.NoError(t, err)
+	
+
+	_, err = io.Copy(f2, reader.CarV1Reader())
 	require.NoError(t, err)
 	require.NoError(t, f2.Close())
 	f3, err := os.Open(f2.Name())
```

Closing; please feel free to re-open if there is anything I might have missed. 🍻

PR #149, opened 2021-07-09
Fix bug in ReadOnly blockstore GetSize

### LINKED ISSUE #133: add a test that verifies if ReadOnly.GetSize actually works
I think its CID reading offset math is wrong.

masih:
Fix a bug where the returned size in `ReadOnly.GetSize` blockstore was
incorrect and the CID was being read from the wrong offset.

Implement tests to assert returned size matches the block raw data size.

Fixes #133


willscott (approved):

masih:
@willscott thank you for the speedy review 🚀

PR #150, opened 2021-07-09
Implement more tests for `ReadOnly` blockstore

masih:
Implement tests that check `ReadOnly` blockstore functions work based on
data payload originated from a V1 or V2 formatted CAR file.

Implement tests that assert instantiation of `ReadOnly` blockstore
fails on unexpected pragma version.

### DIFF
--- a/v2/blockstore/readonly_test.go
+++ b/v2/blockstore/readonly_test.go
@@ -1,43 +1,127 @@
 package blockstore
 
 import (
+	"context"
 	"io"
 	"os"
 	"testing"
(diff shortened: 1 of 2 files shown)


willscott (commented):
maybe also assert that attempting to Put on the read only opened store fails, as does finalize, and that the final file on disk is not changed.

masih:
@willscott thank you for the review; this is ready for another look whenever you get the chance

willscott (approved):

PR #151, opened 2021-07-13
Unexport unecesserry index APIs and integrate with mulitcodec

masih:
Use the codec dedicated to CAR index sorted when marshalling and
unmarshalling `indexSorted`. Note, the code depends on a specific commit
of `go-multicodec` `master` branch. This needs to be replaced with a tag
once a release is made on the go-multicodec side later.

Unexport index APIs that should not be exposed publicly. Remove
`Builder` now that it is not needed anywhere. Move `insertionIndex` into
`blockstore` package since that's the only place it is used.

Introduce an index constructor that takes multicodec code and
instantiates an index.

Fix ignored errors in `indexsorted.go` during marshalling/unmarshlling.

Rename index constructor functions to use consistent terminology; i.e.
`new` instead if `mk`.

Remove redundant TODOs in code.

Relates to:
- https://github.com/multiformats/go-multicodec/pull/46
- #118

### DIFF
--- a/v2/blockstore/insertionindex.go
+++ b/v2/blockstore/insertionindex.go
@@ -1,10 +1,14 @@
-package index
+package blockstore
 
 import (
 	"bytes"
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"io"
+
+	"github.com/ipld/go-car/v2/index"
+	"github.com/multiformats/go-multicodec"
 
 	"github.com/ipfs/go-cid"
 	"github.com/multiformats/go-multihash"
@@ -12,18 +16,21 @@ import (
 	cbor "github.com/whyrusleeping/cbor/go"
 )
 
-type InsertionIndex struct {
-	items llrb.LLRB
-}
+var (
+	errUnsupported      = errors.New("not supported")
+	insertionIndexCodec = 0x300003
+)
 
-func (ii *InsertionIndex) InsertNoReplace(key cid.Cid, n uint64) {
-	ii.items.InsertNoReplace(mkRecordFromCid(key, n))
-}
+type (
+	insertionIndex struct {
+		items llrb.LLRB
+	}
 
-type recordDigest struct {
-	digest []byte
-	Record
-}
+	recordDigest struct {
+		digest []byte
+		index.Record
+	}
+)
 
 func (r recordDigest) Less(than llrb.Item) bool {
 	other, ok := than.(recordDigest)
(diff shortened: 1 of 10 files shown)


v2/blockstore/insertionindex.go
```diff
@@ -1,29 +1,36 @@
...
 	"github.com/petar/GoLLRB/llrb"
 	cbor "github.com/whyrusleeping/cbor/go"
 )
 
-type InsertionIndex struct {
-	items llrb.LLRB
-}
+var (
+	errUnsupported      = errors.New("not supported")
+	insertionIndexCodec = 0x300003
```
mvdan, line 21: surely this can be replaced by go-multicodec now?

v2/index/index.go
```diff
@@ -41,22 +35,22 @@ type (
...
 
-// BuildersByCodec holds known index formats
-// TODO: turn this into a func before the final release?
-var BuildersByCodec = map[Codec]Builder{
-	IndexHashed:       mkHashed,
-	IndexSorted:       mkSorted,
-	IndexSingleSorted: mkSingleSorted,
-	IndexGobHashed:    mkGobHashed,
-	IndexInsertion:    mkInsertion,
+// NewFromCodec constructs a new index corresponding to the given codec.
```
mvdan, line 46: "given codec" is a bit too general IMO, as you only accept the CAR index codecs. How about "to the given CAR index codec"?

v2/index/index.go
```diff
@@ -41,22 +35,22 @@ type (
...
+func NewFromCodec(codec multicodec.Code) (Index, error) {
```
mvdan, line 47: I wonder if just `New` would work. The parameter is a `multicodec.Code`, so it's pretty evident that a codec is used.

And `index.NewFromCodec(multicodec.IndexSorted)` feels repetitive, whereas `index.New(multicodec.IndexSorted)` isn't and still works well.

mvdan (approved):

v2/blockstore/insertionindex.go
masih: this is the insertion index codec. Do we have a codec for that? I thought we only had one for sorted index.

masih: Replaced this specific constant with a wrapped `multicodec.Code`. Please let me know if I missed anything. thanks

mvdan: Ah of course, I had misread the name. SGTM

Issue #145, opened 2021-07-08
Unify utility `Reader`/`Writer` types into `internalio`


masih:
should this live in `internal/io`?

_Originally posted by @willscott in https://github.com/ipld/go-car/pull/144#discussion_r666250418_

masih:
Resolved by https://github.com/ipld/go-car/pull/173

mvdan:
Merged.

PR #154, opened 2021-07-13
Move index generation functionality to root v2 package

masih:
Move index generation functionality to root, because they read/write CAR
files. This leaves the `index` package to only contain the index
implementation itself.

Move `index.Attach` to root while at it following the same logic. Add
TODOs to the implementation of Attach to make it more robust. As is it
_could_ result in corrupt v2 files if offset includes padding since v2
header is not updated in this function.

Add tests to generate index and move existing tests to correspond to the
go file containing generation functionality.


mvdan (approved):
I wonder if this means we can now move parts of internalio to the root package. Not for now, though.

masih:
> I wonder if this means we can now move parts of internalio to the root package. Not for now, though.

Earlier reviews suggested moving more io utilities to internal: https://github.com/ipld/go-car/issues/145

Either works I have no preference. The rule I follow is if it's used in more than one place then move to internal otherwise leave in package that utilises it. What do you think?

Issue #155, opened 2021-07-14
CARv2 blockstore throws "cannot resume from a finalized file"


aarshkshah1992:
After integrating the CARv2 blockstore resumption changes, one of our test fails with the following error:

_"cannot resume from a finalized file"_.

Unfortunately, this is a valid edge case we need to support to support restarts of Filecoin data transfers. Here's the scenario:

1. Storage client transfers a 64Gib file to to the Storage Provider.
2. Storage provider writes the CARv2 file to a  blockstore -> calls `Finalize` but dosen't persist the fact that it's finalized. 
3. Storage provider restarts and attempts to create a CARv2 blockstore from the CARv2 file it has and then finalizes it.

masih:
Fixed by #156

PR #156, opened 2021-07-14
Support resumption in `ReadWrite` blockstore for finalized files

masih:
Support resumption from a complete CARv2 file, effectively implementing
append block to a CARv2 file. Note, on resumption the entire file is
re-indexed. So it is undesirable to do this for complete cars.

Fix a bug where `AllKeysChan` failed on files that contained the index
and resumed from. Fix is done by truncating the CARv2 file, removing the
index if it is present when blockstore is constructed when resuming.
This is because if `AllKeysChan` is called on a file that contains the
index, the ReadOnly backing will continue to read until EOF which will
result in error. Since file is re-indexed on resumption and we require
finalize to be called we truncate the file to remove index.

Clear out CARv2 header on file if resumed from a finalised file to avoid 
issue where the blocks are put and file is closed without finalisation.

Add tests that assert resume with and without finalization. Add tests
that assert read operations work on resumed read-write blockstore.

Fixes: https://github.com/ipld/go-car/issues/155

### DIFF
--- a/v2/blockstore/readwrite.go
+++ b/v2/blockstore/readwrite.go
@@ -41,7 +41,6 @@ type ReadWrite struct {
 	header carv2.Header
 
 	dedupCids bool
-	finalized bool
 }
 
 // TODO consider exposing interfaces
@@ -209,6 +208,11 @@ func (b *ReadWrite) resumeWithRoots(roots []cid.Cid) error {
 			// can't resume.
 			return errors.New("corrupt CARv2 header; cannot resume from file")
 		}
+		// Now that CARv2 header is present on file, clear it to avoid incorrect size and offset in
+		// header in case blocksotre is closed without finalization and is resumed from.
+		if err := b.unfinalize(); err != nil {
+			return err
+		}
 	}
 
 	// Use the given CAR v1 padding to instantiate the CAR v1 reader on file.
@@ -274,8 +278,13 @@ func (b *ReadWrite) resumeWithRoots(roots []cid.Cid) error {
 	return err
 }
 
+func (b *ReadWrite) unfinalize() error {
+	_, err := new(carv2.Header).WriteTo(internalio.NewOffsetWriter(b.f, carv2.PragmaSize))
+	return err
+}
+
 func (b *ReadWrite) panicIfFinalized() {
-	if b.finalized {
+	if b.header.CarV1Size != 0 {
 		panic("must not use a read-write blockstore after finalizing")
 	}
 }
@@ -306,16 +315,14 @@ func (b *ReadWrite) PutMany(blks []blocks.Block) error {
 		}
 		b.idx.insertNoReplace(c, n)
 	}
-	b.header = b.header.WithCarV1Size(uint64(b.carV1Writer.Position()))
-	_, err := b.header.WriteTo(internalio.NewOffsetWriter(b.f, carv2.PragmaSize))
-	return err
+	return nil
 }
 
 // Finalize finalizes this blockstore by writing the CAR v2 header, along with flattened index
 // for more efficient subsequent read.
 // After this call, this blockstore can no longer be used for read or write.
 func (b *ReadWrite) Finalize() error {
-	if b.finalized {
+	if b.header.CarV1Size != 0 {
 		// Allow duplicate Finalize calls, just like Close.
 		// Still error, just like ReadOnly.Close; it should be discarded.
 		return fmt.Errorf("called Finalize twice")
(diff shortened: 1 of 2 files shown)


v2/blockstore/readwrite.go
```diff
@@ -170,14 +174,40 @@ func (b *ReadWrite) resumeWithRoots(roots []cid.Cid) error {
...
+			// are no blocks put unless the user calls finalize.
+			if err := b.f.Truncate(int64(headerInFile.CarV1Offset + headerInFile.CarV1Size)); err != nil {
+				return err
+			}
+		} else {
+			// If CARv1 size is zero, since CARv1 offset wasn't, then the CARv2 header was
+			// most-likely partially written. Since we write the header last in Finalize then the
+			// file most-likely contains the index and we cannot know where it starts, therefore
+			// can't resume.
+			return errors.New("corrupt CARv2 header; cannot resume from file")
```
willscott, line 210: this is unfortunate that partial writes that close uncleanly are now going to be something we aren't comfortable resuming.

It seems more likely that this is a case where the index has not been written at all, right?
It may be worth doing the slower recovery:
* Scan through the blocks in the carV1
* For each, see if the length is valid with regards to the length of the file
* either re-hash each block and make sure the contents match the CID, or ideally build up the index while doing the scan of lengths, and then do that hashing backwards until you find a valid block.
* if there is content after the last valid block, and it parses as an index, great
* if there is at most 1 final invalid block, we crashed in the middle of writing it and can truncate to before that and resume.
* otherwise we have an actual corrupt car and should force a new full export.

v2/blockstore/readwrite_test.go
```diff
@@ -291,25 +291,62 @@ func TestBlockstoreResumption(t *testing.T) {
...
-			// Note, we don't have to close the file for resumption to work.
-			// We do this to avoid resource leak during testing.
-			require.NoError(t, subject.Close())
+		wantBlockCountSoFar++
+		wantBlocks[b.Cid()] = b
+
+		// 30% chance of subject failing; more concretely: re-instantiating blockstore with the same
+		// file without calling Finalize. The higher this percentage the slower the test runs
+		// considering the number of blocks in the original CAR v1 test payload.
+		resume := rand.Float32() <= 0.3
```
willscott, line 308: we should make tests deterministic

willscott: @mvdan does go use a fixed seed for rand by default in these test cases?

masih: Would a `rand` with fixed seed suffice?

masih: > @mvdan does go use a fixed seed for rand by default in these test cases?

I _think_ this is a yes.

v2/blockstore/readwrite.go
masih: > It seems more likely that this is a case where the index has not been written at all, right?

No, I think it's the case that the header is actually corrupt, since I have changed the code to write V2 header with the correct CarV1 size on every put. The rationale for doing this is in the commit message. TLDR; `Put` on finalised files that then closed without finalisation means header looks complete but actually missing latest put blocks.

On the slow recovery items listed above, I think that's useful regardless in recovering from files with tailing corrupt blocks.

willscott: doing a write to the header on every put is potentially going to be a much worse write pattern for disk - instead of a linear write that an HDD would be happy with, we'll be jumping back and forth in our access pattern, which is going to potentially result in significantly slower performance.

masih: Agreed... It did not feel good writing it. What alternative can you think of?

willscott: could we keep the carv1size as "-1" (or umaxint if it's unsigned) to indicate "until end of file"?

masih: We could but the issue is resuming from a valid and finalised CARv2 file. That file will contain the index and a complete header.

I guess what we could do is:
* if resuming from a file with populated header (i.e. non-zero carv1 size and offset) then remove index by truncating the file and _wipe off the valid header_  so that further `put`s without finalisation forces future resumptions to read file to the end.

v2/blockstore/readwrite_test.go
mvdan: This code is using the global source, so you can't know for sure. It's likely that some dependency somewhere is doing something akin to `func init() { rand.Seed(...) }`.

So yeah, use your custom source with a fixed seed.

v2/blockstore/readwrite.go
masih: @willscott capture the workflow you kindly listed above in an improvement issue for another PR here: https://github.com/ipld/go-car/issues/157 Thank you for pointing this out

masih:
@willscott  thank you for the review; this is ready for another look whenever you get the chance 🍻

willscott (approved):

PR #158, opened 2021-07-15
Index error found should return block not found

aarshkshah1992:
There's blockstore dependent code in our libraries that depends on this particular error if a block does NOT exist.


masih (approved):
Thank you for pointing this out

Issue #104, opened 2021-06-24
Implement a CARv2 SelectiveCARAPI when clients have upgraded to go-ipld-prime v0.9.0


aarshkshah1992:
https://github.com/ipld/go-car/pull/102 removes the CARv1 dependency from the WIP CARv2 library  and states the motivation for doing so i.e. to not force clients to upgrade to the breaking `go-ipld-prime v0.9.0` release because the latest CARv1 release depends on `go-ipld-prime v0.9.0`.

Towards that goal, the CARv2 library also holds back from implementing a `SelectiveCAR` API for v2 as the `SelectiveCAR` implementation would need `go-ipld-prime` and we should be using the latest ipld prime release i.e. `go-ipld-prime v0.9.0` to implement it rather than using an older IPLD prime release.

This issue is meant to capture the fact that we should implementing a CARv2 `SelectiveCAR` API when clients have upgraded to `go-ipld-prime v0.9.0`.

masih:
Fixed by https://github.com/ipld/go-car/pull/269

PR #160, opened 2021-07-15
Remove dependency to `bufio.Reader` in internal `carv1` package

masih:
Remove dependency to `bufio.Reader` in internal `carv1` package that
seems to be mainly used for peeking a byte to return appropriate error
when stream abruptly ends, relating to #36. This allows simplification
of code across the repo and remove all unnecessary wrappings of
`io.Reader` with `bufio.Reader`. This will also aid simplify the
internal IO utilities which will be done in future PRs. For now we
simply remove dependency to `bufio.Reader`.

Relates to:
- https://github.com/ipld/go-car/pull/36

### DIFF
--- a/v2/blockstore/readonly.go
+++ b/v2/blockstore/readonly.go
@@ -1,7 +1,6 @@
 package blockstore
 
 import (
-	"bufio"
 	"bytes"
 	"context"
 	"errors"
@@ -128,7 +127,7 @@ func OpenReadOnly(path string) (*ReadOnly, error) {
 }
 
 func (b *ReadOnly) readBlock(idx int64) (cid.Cid, []byte, error) {
-	bcid, data, err := util.ReadNode(bufio.NewReader(internalio.NewOffsetReadSeeker(b.backing, idx)))
+	bcid, data, err := util.ReadNode(internalio.NewOffsetReadSeeker(b.backing, idx))
 	return bcid, data, err
 }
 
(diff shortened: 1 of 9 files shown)


rvagg (approved):
oh, so we have a carv1 inside carv2? is that going to be deduped at some point?

I have some memory of this `Peek()` code, trying to make it give proper errors on EOF vs "can't decode this uvarint". Other than that, I don't see a good reason to hold on to bufio.Reader; so go ahead and yeet it! This looks good to me. I'd be happy to see this migrate into the top level package too, I don't think there's a strong reason to keep it otherwise since that `Peek()` is not protecting an unbounded reader or anything.

masih:
> oh, so we have a carv1 inside carv2? is that going to be deduped at some point?

Yep the context behind this is captured in https://github.com/ipld/go-car/issues/104. TLDR; `go-ipld-prime` dependency in FileCoin side needs to be upgraded before we can deduplicate. The plan is certainly to deduplicate. The same dependency issue on FileCoin side is blocking the implementation of Selective Car in v2. Forking the v1 into v2 also allows us to do such refactors.


>  I'd be happy to see this migrate into the top level package too
Yeah that would be nice; the bad thing is we can't roll out breaking changes so we have to leave the `buffo.Reader` taking APIs in v1 in place. But we can certainly mark them as deprecated and roll out new versions. Not sure how nice the naming of those functions would be since they are named things like `ReadHeader`.

mvdan:
SGTM. For what it's worth I'm not convinced we can simply deduplicate the carv1 packages; the v0 ones in master have APIs that limit what we can do, like `func ReadHeader(br *bufio.Reader) (*CarHeader, error)` still requiring a buffered reader.

We can certainly try to deduplicate some of it though.

rvagg:
> the bad thing is we can't roll out breaking changes

.. but isn't `bufio.Reader` a superset of `io.Reader`? if we switched to just `r *io.Reader` wouldn't it still work if someone passed in a `bufio.Reader`?

masih:
> > the bad thing is we can't roll out breaking changes
> 
> .. but isn't `bufio.Reader` a superset of `io.Reader`? if we switched to just `r *io.Reader` wouldn't it still work if someone passed in a `bufio.Reader`?

mm good point; is that OK as far as Go API etiquette goes? @mvdan  what do you think?

mvdan:
It's technically a breaking change, as one could e.g. pass ReadHeader as a parameter to a func that expects the bufio type.

It's likely that it would break noone in practice though, so we could always try and see. It's a v0 after all. I haven't checked if any other bits of the API needed changing for their uses in carv2.

PR #162, opened 2021-07-15
Enable deduplication by CID in `ReadWrite` blockstore by default

masih:
Enable the deduplication by CID in `ReadWrite` blockstore by default,
since it is the typical and common usecase across current usage
patterns.

### DIFF
--- a/v2/blockstore/readwrite.go
+++ b/v2/blockstore/readwrite.go
@@ -59,13 +59,15 @@ func WithIndexPadding(p uint64) Option {
 	}
 }
 
-// WithCidDeduplication makes Put calls ignore blocks if the blockstore already
+// WithDisableCidDeduplication allows Put calls to write blocks even though blockstore already
 // has the exact same CID.
-// This can help avoid redundancy in a CARv1's list of CID-Block pairs.
+//
+// Note, deduplication by CID is enabled by default, which can help avoid redundancy in a CARv1's list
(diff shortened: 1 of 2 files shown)


v2/blockstore/readwrite.go
```diff
@@ -59,13 +59,15 @@ func WithIndexPadding(p uint64) Option {
...
 // has the exact same CID.
-// This can help avoid redundancy in a CARv1's list of CID-Block pairs.
+//
+// Note, deduplication by CID is enabled by default, which can help avoid redundancy in a CARv1's list
+// of CID-Block pairs.
 //
 // Note that this compares whole CIDs, not just multihashes.
-func WithCidDeduplication(b *ReadWrite) { // TODO should this take a bool and return an option to allow disabling dedupliation?
-	b.dedupCids = true
+func WithDisableCidDeduplication(b *ReadWrite) {
```
willscott, line 69: this is a weird side-effect-y setup. Is there a way to only allow this as an option at creation, rather than toggling it during usage?

masih: Good point. for this to happen I _think_ this needs to return an option.

mvdan: I'm working on unifying the options APIs now, which will only be accepted at the creation points like NewFoo and OpenFoo.

PR #165, opened 2021-07-15
Improve test coverage for `ReadWrite` blockstore

masih:
Add tests that assert:
 - when padding options are set the payload is as
expected
 - when finalized, blockstore calls panic
 - when resumed from mismatching padding error is as expected
 - when resumed from non-v2 file error is as expected

Remove redundant TODOs in code

### DIFF
--- a/v2/blockstore/readwrite_test.go
+++ b/v2/blockstore/readwrite_test.go
@@ -12,6 +12,8 @@ import (
 	"testing"
 	"time"
 
+	dag "github.com/ipfs/go-merkledag"
+
 	carv2 "github.com/ipld/go-car/v2"
 	"github.com/ipld/go-car/v2/index"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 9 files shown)


v2/blockstore/readwrite_test.go
```diff
@@ -12,6 +12,8 @@ import (
 	"testing"
 	"time"
 
+	dag "github.com/ipfs/go-merkledag"
```
mvdan, line 15: I think merkledag would be better FYI, for clarity

v2/blockstore/readwrite_test.go
```diff
@@ -26,7 +28,11 @@ import (
 	"github.com/ipld/go-car/v2/internal/carv1"
 )
 
-var rng = rand.New(rand.NewSource(1413))
+var (
+	rng              = rand.New(rand.NewSource(1413))
+	oneTestBlock     = dag.NewRawNode([]byte("fish")).Block
+	anotherTestBlock = dag.NewRawNode([]byte("barreleye")).Block
```
mvdan, line 34: I was going to suggest to use https://pkg.go.dev/github.com/ipfs/go-block-format#NewBlock instead, but then I noticed that uses a cidv0 instead of merkledag's cidv1.

mvdan (approved):

v2/blockstore/readwrite_test.go
masih: Interesting; we probably should use both actually. thanks for pointing this out

masih: Added both thanks @mvdan

v2/blockstore/readwrite_test.go
masih: `dag` was used consistently across tests; replaced all with `merkledag`.

PR #169, opened 2021-07-16
CidReader panics for a valid cid

aarshkshah1992:
See Test. Needed for M1.


mvdan:
Being fixed in https://github.com/ipfs/go-cid/pull/128. I'll close this when that's merged and we bump the dep.

mvdan:
Fixed by https://github.com/ipld/go-car/pull/170.

PR #172, opened 2021-07-16
Implement examples and tests for `index` package

masih:
Add examples that show how to read and write an index to/from file.

Test marshalling and unmarshalling index files.


v2/index/example_test.go
```diff
@@ -0,0 +1,95 @@
...
+	defer cr.Close()
+
+	// Get root CIDs in the CARv1 file.
+	roots, err := cr.Roots()
+	if err != nil {
+		panic(err)
+	}
+
+	// Read and unmarshall index within CARv2 file.
+	ir := cr.IndexReader()
```
mvdan, line 29: inline this var?

v2/index/example_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+	// Read and unmarshall the destination file as a separate index instance.
+	reReadIdx, err := index.ReadFrom(f)
+	if err != nil {
+		panic(err)
+	}
+
+	// Expect indices to be equal.
+	if reflect.DeepEqual(idx, reReadIdx) {
+		fmt.Printf("Saved index file at %v  matches the index embedded in CARv2 at %v.\n", dest, src)
```
mvdan, line 88: nit: double space

v2/index/example_test.go
```diff
@@ -0,0 +1,95 @@
...
+	// Open the CARv2 file
+	src := "../testdata/sample-wrapped-v2.car"
+	cr, err := carv2.OpenReader(src)
+	if err != nil {
+		panic(err)
+	}
+	defer cr.Close()
+
+	// Read and unmarshall index within CARv2 file.
+	ir := cr.IndexReader()
```
mvdan, line 60: ditto

v2/index/index_test.go
```diff
@@ -0,0 +1,177 @@
...
+			"IndexHashedMultiCodecIsError",
+			multicodec.Code(indexHashed),
+			nil,
+			true,
+		},
+		{
+			"IndexGobHashedMultiCodecIsError",
+			multicodec.Code(indexGobHashed),
+			nil,
+			true,
```
mvdan, line 53: please use explicit keys for these; otherwise I just see `nil, true,` and I have no clue what they mean without scrolling back to the top.

not a huge problem here given the few tests, but it's good to be consistently specific for table-driven tests, for clarity - and because we might add more tests later :)

v2/index/index_test.go
```diff
@@ -0,0 +1,177 @@
...
+	idxf, err := os.Open("../testdata/sample-index.carindex")
+	require.NoError(t, err)
+	t.Cleanup(func() { idxf.Close() })
+
+	// Unmarshall to get expected index
+	wantIdx, err := ReadFrom(idxf)
+	require.NoError(t, err)
+
+	// Save the same index at destination
+	dest := path.Join(t.TempDir(), "index-write-to-test.carindex")
```
mvdan, line 147: remember to always use filpath.Join for paths on disk

v2/index/index_test.go
```diff
@@ -0,0 +1,177 @@
...
+	require.NoError(t, Save(wantIdx, dest))
+
+	// Open the saved file
+	destF, err := os.Open(dest)
+	require.NoError(t, err)
+	t.Cleanup(func() { destF.Close() })
```
mvdan, line 153: nit: we should be checking close errors for files that we write to. not hugely important for a test, but still...

mvdan (approved):

PR #176, opened 2021-07-16
Rename `CarV1` with `Data` in API to be consistent with spec

masih:
Rename terminology to match what the spec uses to describe the inner
CARv1 payload, i.e. Data payload.

Update docs to use CARv1 and CARv2 consistently.

Fix typo in Options API.

See:
- https://ipld.io/specs/transport/car/carv2/

### DIFF
--- a/v2/car.go
+++ b/v2/car.go
@@ -33,7 +33,7 @@ type (
 		DataOffset uint64
 		// The byte-length of the CARv1 data payload.
 		DataSize uint64
-		// Te byte-offset from the beginning of the CARv2 to the first byte of the index payload. This value may be 0 to indicate the absence of index data.
+		// The byte-offset from the beginning of the CARv2 to the first byte of the index payload. This value may be 0 to indicate the absence of index data.
 		IndexOffset uint64
 	}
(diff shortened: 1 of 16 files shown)


v2/car.go
```diff
@@ -25,18 +25,18 @@ var Pragma = []byte{
...
-		// The offset from the beginning of the file at which the dump of CAR v1 starts.
-		CarV1Offset uint64
-		// The size of CAR v1 encapsulated in this CAR v2 as bytes.
-		CarV1Size uint64
-		// The offset from the beginning of the file at which the CAR v2 index begins.
+		// The byte-offset from the beginning of the CARv2 to the first byte of the CARv1 data payload.
+		DataOffset uint64
+		// The byte-length of the CARv1 data payload.
+		DataSize uint64
+		// Te byte-offset from the beginning of the CARv2 to the first byte of the index payload. This value may be 0 to indicate the absence of index data.
```
mvdan, line 36:
```suggestion
		// The byte-offset from the beginning of the CARv2 to the first byte of the index payload. This value may be 0 to indicate the absence of index data.
```

mvdan (approved):

Issue #124, opened 2021-06-29
Godoc with example usage of carv2 primary flows


willscott:
There should be useful documentation and examples guiding users through

* [x] an import/export roundtrip
* [x] Opening an unknown file as a blockstore
* [x] writing data to a carv2 via blockstore interface and saving it.
* [ ] Top level doc todo

willscott:
make sure also to fix the todo in top level godoc:
https://github.com/ipld/go-car/blob/master/v2/doc.go#L2

mvdan:
Yep! I actually noticed this just minutes before you wrote this comment, too, as I was looking at the pkgsite links I shared :)

PR #180, opened 2021-07-16
Add example blockstore examples

masih:
Implement examples that show how to open a read-only blockstore, and a
read-write blockstore along with resumption from the same file.

Fixes:
- https://github.com/ipld/go-car/issues/124


v2/blockstore/example_test.go
```diff
@@ -0,0 +1,146 @@
...
+
+	// Print the raw data size for the first 5 CIDs in the CAR file.
+	keysChan, err := robs.AllKeysChan(context.Background())
+	if err != nil {
+		panic(err)
+	}
+	fmt.Println("List of first 5 CIDs and their raw data size:")
+	var i int
+	for k := range keysChan {
+		if i > 4 {
```
mvdan, line 49: if you want to print 5, I think it's then easier to do `i >= 5`, to use the same number everywhere. you might as well make that a const/var and use it in the print above

v2/blockstore/example_test.go
```diff
@@ -0,0 +1,146 @@
...
+	// Blocks put after resumption are also present.
+	block, err = resumedRwbos.Get(andTheOtherBlock.Cid())
+	if err != nil {
+		panic(err)
+	}
+	fmt.Printf("It also contains the block put after resumption with raw value of `%v`.\n", string(block.RawData()))
+
+	// Finalize the blockstore to flush out the index and make a complete CARv2.
+	// Note, Finalize must be called on an open ReadWrite blockstore to flush out a complete CARv2.
+	if err = resumedRwbos.Finalize(); err != nil {
```
mvdan, line 137: nit: this can be `:=`; it helps keep the code simpler to follow as you're not modifying the var in the parent scope. this can be done for any other place where you only assign an error.

mvdan (approved):

PR #182, opened 2021-07-16
Remove `index.Save` API

masih:
It's two lines of code extra and we already have read from and write to
APIs for index.

Remove `hydrate` cli since the assumptions it makes about appending
index are no longer true. header needs updating for example. Removing to
be re-added when needed as part of a propper CARv2 CLI.


v2/index/example_test.go
```diff
@@ -63,17 +64,21 @@ func ExampleSave() {
 
 	// Store the index alone onto destination file.
 	dest := "../testdata/sample-index.carindex"
-	err = index.Save(idx, dest)
+	f, err := os.Create(dest)
+	if err != nil {
+		panic(err)
+	}
+	defer f.Close()
```
mvdan, line 71: since this is an example, please check its error at the end of the func, too :)

mvdan (approved):

PR #183, opened 2021-07-20
blockstore: implement UseWholeCIDs

mvdan:
(see commit message)


mvdan:
FYI @masih, there are some open questions here:

1) I think we should remove the non-specced and unused index implementations. They are slowing us down and they're unused. They can always be recovered from the git history. For now, I've made their GetAll panic.

2) Should we remove Index.Get and have the blockstore always call Index.GetAll? It already does that for Has/Get/GetSize, to deduplicate code between "whole cid" and "multihash" modes. It's only Put that does an index Get, when deduplicating puts by multihash, but that could be a GetMany that stops the iteration at the first callback.

mvdan:
Re 2, I lean towards removing Get. GetAll is more generic, as it allows supporting duplicates and the "whole CID" lookups. Perhaps there's an argument that GetAll is more costly with its callback, but I doubt that will matter in practice - and if it does, we could always try to optimize it, or fall back to re-adding Get.

v2/blockstore/readonly.go
```diff
@@ -177,68 +176,121 @@ func (b *ReadOnly) Has(key cid.Cid) (bool, error) {
...
+			}
+			return false
+		}
+	})
+	if errors.Is(err, index.ErrNotFound) {
 		return -1, blockstore.ErrNotFound
+	} else if err != nil {
+		return -1, err
+	} else if fnErr != nil {
+		return -1, fnErr
```
masih, line 288: non blocker: at some point it would be nice to refactor duplicate logic in iterating through `index.GetAll` relative to options.

v2/options.go
```diff
@@ -46,8 +46,10 @@ type ReadWriteOption interface {
 // a zero-length section as the end of the input CAR file. For example, this can
 // be useful to allow "null padding" after a CARv1 without knowing where the
 // padding begins.
-func ZeroLengthSectionAsEOF(o *ReadOptions) {
-	o.ZeroLengthSectionAsEOF = true
+func ZeroLengthSectionAsEOF(enable bool) ReadOption {
+	return func(o *ReadOptions) {
+		o.ZeroLengthSectionAsEOF = true
```
masih, line 51: Thinking out-loud are there any other options in code that we want to make configurable like this before the release tag?

masih (approved):
Non-blocker comments:
* Would be great to refactor logic of dealing with `index.GetAll` in both blockstore and index.
* Now that the whole CID option is implemented it would be awesome to expand tests both for blockstore and index `GetAll`.



> I think we should remove the non-specced and unused index implementations.

That makes sense; delete them indices! 🔫 

> Should we remove Index.Get and have the blockstore always call Index.GetAll? It already does that for Has/Get/GetSize, to deduplicate code between "whole cid" and "multihash" modes. It's only Put that does an index Get, when deduplicating puts by multihash, but that could be a GetMany that stops the iteration at the first callback.

I think from consistency perspective alone this makes makes sense to do. 
In parallel to this, it would be fantastic to measure how `index.GetAll` would impact performance. Regardless, for me API consistency is more important.

Nice one @mvdan 🍻

masih:
> Re 2, I lean towards removing Get. GetAll is more generic, as it allows supporting duplicates and the "whole CID" lookups. Perhaps there's an argument that GetAll is more costly with its callback, but I doubt that will matter in practice - and if it does, we could always try to optimize it, or fall back to re-adding Get.

I agree. I think any existing usage of `Get` can be implemented with `GetAll`.
Would be awesome to measure the performance penalty.

v2/options.go
mvdan: I didn't find any. I'm honestly not too bothered about boolean vs "just let the user set the non-default", but either way, I wanted it to be consistent.

mvdan:
Re perf: Raul already asked for baseline benchmarks, so I'll do it as part of that, and separate from this PR. I definitely agree that correctness comes first, and making all blockstore methods consistently obey UseWholeCIDs is best.

v2/blockstore/readonly.go
mvdan: Maybe, I was thinking about it too, though note that their behavior on "none found" is slightly different - only some return ErrNotFound.

masih: We can always refactor this later; for now this is all good.

mvdan:
Applied both suggestions mentioned earlier, as discussed - just clarifying in case someone else sees this PR.

PR #184, opened 2021-07-20
Avoid writing to files in `testdata`

### LINKED ISSUE #175: CARv2 TestBlockstore fails intermittently
Happened on Windows CI build [here](https://github.com/ipld/go-car/runs/3086058640?check_suite_focus=true). 

```
      readwrite_test.go:111: AllKeysChan returned an unexpected amount of keys; expected 1049 but got 0
123
```

masih:
Avoid writing to files in `testdata` through examples. This is because
when tests that use those file are run in parallel will fail if files
are modified. Instead write to a temporary file in examples, and
whenever opening testdata files in `RW` mode, make a temporary copy.

Thanks to @mvdan for pointing this out.

Fixes #175

### DIFF
--- a/v2/index/example_test.go
+++ b/v2/index/example_test.go
@@ -3,7 +3,9 @@ package index_test
 import (
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
+	"path/filepath"
 	"reflect"
 
 	carv2 "github.com/ipld/go-car/v2"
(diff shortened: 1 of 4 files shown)


masih:
`stress`ed tests in all packages for a minute locally and all seems fine.

v2/index/example_test.go
```diff
@@ -67,8 +69,12 @@ func ExampleWriteTo() {
...
 
 	// Store the index alone onto destination file.
-	dest := "../testdata/sample-index.carindex"
-	f, err := os.Create(dest)
+	tdir, err := ioutil.TempDir(os.TempDir(), "example-*")
+	if err != nil {
+		panic(err)
+	}
+	dst := filepath.Join(tdir, "index.carindex")
+	f, err := os.Create(dst)
```
mvdan, line 77: just use ioutil.TempFile directly?

mvdan (approved):

PR #185, opened 2021-07-21
Allow `ReadOption`s to be set when getting or generating index

masih:
Fix a bug in read options where the value of zero-len option is always
set to `true` even if the flag passed is false. This meant if the option
was present it would have always enabled the zero-len as EOF.

Allow the user to specify read options on `ReadOrGenerateIndex`. Pass
the options onto downstream reader/generators.

Generate a CARv1 file with null padding and check it into `testdata`. In
addition check in another such file from ignite just to have two
different samples of this case.

Enhance index generation tests to expect error when zero-len option is
not set, and generate index when it is for files with null padding.

Make RW blockstore test use the fixed test files from `testdata` instead
of generating every time.


v2/options.go
```diff
@@ -48,7 +48,7 @@ type ReadWriteOption interface {
 // padding begins.
 func ZeroLengthSectionAsEOF(enable bool) ReadOption {
 	return func(o *ReadOptions) {
-		o.ZeroLengthSectionAsEOF = true
+		o.ZeroLengthSectionAsEOF = enable
```
aarshkshah1992, line 51: nice catch :)

aarshkshah1992 (approved):
lgtm.

v2/options.go
mvdan: 🤦

mvdan (approved):

v2/options.go
masih: To clarify this did not cause the issue; it just meant if the option is present it's always enabled. Totally my fault for missing it in the PR review.

aarshkshah1992: Yeah ofcourse, I agree this wasn't the RC.

v2/index_gen_test.go
```diff
@@ -56,7 +95,7 @@ func TestReadOrGenerateIndex(t *testing.T) {
 			carFile, err := os.Open(tt.carPath)
 			require.NoError(t, err)
 			t.Cleanup(func() { assert.NoError(t, carFile.Close()) })
-			got, err := ReadOrGenerateIndex(carFile)
+			got, err := ReadOrGenerateIndex(carFile, tt.readOpts...)
 			if tt.wantErr {
```
dirkmc, line 99: Should there be an else case here that does `require.NoError(t, err)`?

dirkmc (approved):

v2/index_gen_test.go
masih: Does this not suffice?
```
if tt.wantErr {
  require.Error(t, err)
}
```

masih: Doh; ignore me; thank you @dirkmc will add 🍻

dirkmc: I'm saying:
```
if tt.wantErr {
  require.Error(t, err)
} else {
  require.NoError(t, err)
}
```

So that it verifies that error is nil if we're _not_ expecting an error

masih: Fixed here: https://github.com/ipld/go-car/pull/186 Thank you for pointing this out 🍻

Issue #112, opened 2021-06-28
v2: consider adding logging options


mvdan:
For example:

```
func WithLogging(func(format string, args ...interface{})) Option
```

See https://github.com/ipld/go-car/pull/111#discussion_r659582951. There are a few places where we can't easily return errors, such as AllKeysChan's inner goroutine, so it's a shame if the best we can do is simply throw them away.

cc @dirkmc

mvdan:
Removing from the milestone for now; we agreed this is not a pressing need for Ignite, and it can always be added later.

mvdan:
After talking to Masih, we agreed that opt-in logging is still good, but error reporting via per-blockstore logging isn't great. So errors from AllKeysChan is now https://github.com/ipld/go-car/issues/177.

BigLep:
Good idea.  Closing for now since no usecase is driving this.  We can reopen if/when that's the case.

PR #187, opened 2021-07-23
Propagate async `blockstore.AllKeysChan` errors via context

### LINKED ISSUE #177: v2: allow AllKeysChan to report errors via a callback on Context
The API would add:

```
var ctxKeyErrorHandler someUnexportedType

func WithErrorHandler(ctx context.Context, func(error)) context.Context
```

And then one can specify an error handler for AllKeysChan:

```
ctx := someParentContext
ctx = blockstore.WithErrorHandler(ctx, func(err error) {
    // handle the error
})
... = bs.AllKeysChan(ctx, ...)
```

Separate from https://github.com/ipld/go-car/issues/112, since that's logging, and error values are lost.

masih:
Implement a mechanism that allows the user to hook an error handling
function to the context passed to `AllKeysChan`. The function is then
notified when an error occurs during asynchronous traversal of data
payload.

Add tests that assert the success and failure cases when error handler
is set.

Fixes #177

### DIFF
--- a/v2/blockstore/readonly.go
+++ b/v2/blockstore/readonly.go
@@ -24,26 +24,34 @@ import (
 
 var _ blockstore.Blockstore = (*ReadOnly)(nil)
 
-// ReadOnly provides a read-only CAR Block Store.
-type ReadOnly struct {
-	// mu allows ReadWrite to be safe for concurrent use.
-	// It's in ReadOnly so that read operations also grab read locks,
-	// given that ReadWrite embeds ReadOnly for methods like Get and Has.
-	//
-	// The main fields guarded by the mutex are the index and the underlying writers.
-	// For simplicity, the entirety of the blockstore methods grab the mutex.
-	mu sync.RWMutex
+var errZeroLengthSection = fmt.Errorf("zero-length section not allowed by default; see WithZeroLengthSectionAsEOF option")
 
-	// The backing containing the data payload in CARv1 format.
-	backing io.ReaderAt
-	// The CARv1 content index.
-	idx index.Index
+type (
+	// ReadOnly provides a read-only CAR Block Store.
+	ReadOnly struct {
+		// mu allows ReadWrite to be safe for concurrent use.
+		// It's in ReadOnly so that read operations also grab read locks,
+		// given that ReadWrite embeds ReadOnly for methods like Get and Has.
+		//
+		// The main fields guarded by the mutex are the index and the underlying writers.
+		// For simplicity, the entirety of the blockstore methods grab the mutex.
+		mu sync.RWMutex
 
-	// If we called carv2.NewReaderMmap, remember to close it too.
-	carv2Closer io.Closer
+		// The backing containing the data payload in CARv1 format.
+		backing io.ReaderAt
+		// The CARv1 content index.
+		idx index.Index
 
-	ropts carv2.ReadOptions
-}
+		// If we called carv2.NewReaderMmap, remember to close it too.
+		carv2Closer io.Closer
+
+		ropts carv2.ReadOptions
+	}
+
+	contextKey string
+)
+
+const asyncErrHandlerKey contextKey = "asyncErrorHandlerKey"
 
 // UseWholeCIDs is a read option which makes a CAR blockstore identify blocks by
 // whole CIDs, and not just their multihashes. The default is to use
@@ -303,7 +311,19 @@ func (b *ReadOnly) PutMany([]blocks.Block) error {
 	panic("called write method on a read-only blockstore")
 }
 
-// AllKeysChan returns the list of keys in the CAR.
+// WithAsyncErrorHandler returns a context with async error handling set to the given errHandler.
+// Any errors that occur during asynchronous operations of AllKeysChan will be passed to the given
+// handler.
+func WithAsyncErrorHandler(ctx context.Context, errHandler func(error)) context.Context {
+	return context.WithValue(ctx, asyncErrHandlerKey, errHandler)
+}
+
+// AllKeysChan returns the list of keys in the CAR data payload.
+// If the ctx is constructed using WithAsyncErrorHandler any errors that occur during asynchronous
+// retrieval of CIDs will be passed to the error handler function set in context.
+// Otherwise, errors will terminate the asynchronous operation silently.
+//
+// See WithAsyncErrorHandler
 func (b *ReadOnly) AllKeysChan(ctx context.Context) (<-chan cid.Cid, error) {
 	// We release the lock when the channel-sending goroutine stops.
 	b.mu.RLock()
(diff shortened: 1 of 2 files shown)


v2/blockstore/readonly.go
```diff
@@ -22,28 +22,38 @@ import (
...
-
-	// If we called carv2.NewReaderMmap, remember to close it too.
-	carv2Closer io.Closer
-
-	ropts carv2.ReadOptions
-}
+var (
+	_                    blockstore.Blockstore = (*ReadOnly)(nil)
+	noopErrHandler                             = func(error) {}
+	errZeroLengthSection                       = fmt.Errorf("zero-length section not allowed by default; see WithZeroLengthSectionAsEOF option")
```
mvdan, line 28: IMHO this grouping isn't great, as it adds a lot of aligning whitespace, and now the Errorf barely fits on most screens.

I'd add an empty line in between the first and second declarations, or just drop the grouping, as it's not gaining you much.

v2/blockstore/readonly.go
```diff
@@ -22,28 +22,38 @@ import (
...
+		// If we called carv2.NewReaderMmap, remember to close it too.
+		carv2Closer io.Closer
+
+		ropts carv2.ReadOptions
+	}
+
+	contextKey int
+)
+
+const asyncErrHandlerKey contextKey = iota
```
mvdan, line 56: why use iota at all, if this is just one value?

v2/blockstore/readonly.go
```diff
@@ -404,3 +434,13 @@ func (b *ReadOnly) closeWithoutMutex() error {
 	}
 	return nil
 }
+
+// asyncErrHandlerFrom gets an error handler from ctx by checking if a value is associated with key
+// asyncErrHandlerKey. If no such value is present then it will return noopErrHandler.
+func asyncErrHandlerFrom(ctx context.Context) func(error) {
```
mvdan, line 440: this should be right next to the func that sets the context value, for clarity

v2/blockstore/readonly.go
```diff
@@ -22,28 +22,38 @@ import (
...
-	idx index.Index
-
-	// If we called carv2.NewReaderMmap, remember to close it too.
-	carv2Closer io.Closer
-
-	ropts carv2.ReadOptions
-}
+var (
+	_                    blockstore.Blockstore = (*ReadOnly)(nil)
+	noopErrHandler                             = func(error) {}
```
mvdan, line 27: this should be a proper func declaration, not a func variable. func variables should only be used when the function can be nil or its value isn't static, but neither applies here.

v2/blockstore/readonly.go
```diff
@@ -364,8 +394,8 @@ func (b *ReadOnly) AllKeysChan(ctx context.Context) (<-chan cid.Cid, error) {
 			select {
 			case ch <- c:
 			case <-ctx.Done():
-				// TODO: log ctx error
-				return
+				errHandler(ctx.Err())
+				return // TODO: log ctx error
```
mvdan, line 398: I think all these error TODOs can be removed now; we allow the user to handle them however they want via the error handlers.

v2/blockstore/readonly.go
```diff
@@ -404,3 +434,13 @@ func (b *ReadOnly) closeWithoutMutex() error {
 	}
 	return nil
 }
+
+// asyncErrHandlerFrom gets an error handler from ctx by checking if a value is associated with key
+// asyncErrHandlerKey. If no such value is present then it will return noopErrHandler.
+func asyncErrHandlerFrom(ctx context.Context) func(error) {
+	value := ctx.Value(asyncErrHandlerKey)
+	if eh, ok := value.(func(error)); ok {
```
mvdan, line 442: nit: check for `eh != nil` instead of `ok`, because otherwise you can't guarantee that this func never returns nil. otherwise, I can supply a nil func to WithAsyncErrorHandler and make AllKeysChan panic.

the alternative is to make AllKeysChan check `errHandler != nil` before each call, in which case you would also just `return nil` in the fallback instead of using a noop func. I'd probably go that route myself, because you also gain the ability to easily tell if an error handler is being used or not.

v2/blockstore/readonly_test.go
```diff
@@ -118,6 +118,84 @@ func TestNewReadOnlyFailsOnUnknownVersion(t *testing.T) {
...
+		defer close(errHandlerCalled)
+		require.EqualError(t, err, "context deadline exceeded")
+	})
+	_, err = subject.AllKeysChan(expiredErrHandlingCtx)
+	require.NoError(t, err)
+
+	// Assert error handler was called with required condition, waiting at most 3 seconds.
+	select {
+	case <-errHandlerCalled:
+		t.Log("error handler called as expected")
```
mvdan, line 141: I don't think tests should log when things work normally, unless the test is very slow to the point of needing to show progress :) Otherwise `go test -v` would tend to be very verbose

mvdan (approved):

v2/blockstore/readonly.go
masih: `gofumpt` for you.

v2/blockstore/readonly.go
masih: The rationale was: it is one value today. Don't know if we would add to this or not in the future. Will make it zero.

v2/blockstore/readonly.go
masih: I am curious what algorithm should I follow for grouping if `gofumpt` is not always perfect?

v2/blockstore/readonly.go
masih: Left them since we have a dedicated issue on logging captured in https://github.com/ipld/go-car/issues/112

v2/blockstore/readonly.go
```diff
@@ -303,11 +313,26 @@ func (b *ReadOnly) PutMany([]blocks.Block) error {
...
+// retrieval of CIDs will be passed to the error handler function set in context.
+// Otherwise, errors will terminate the asynchronous operation silently.
+//
+// See WithAsyncErrorHandler
 func (b *ReadOnly) AllKeysChan(ctx context.Context) (<-chan cid.Cid, error) {
 	// We release the lock when the channel-sending goroutine stops.
 	b.mu.RLock()
 
+	// Get async error handler from the context.
+	errHandler := asyncErrHandlerFrom(ctx)
```
iand, line 334: A drive-by comment: since all uses of errHandler are immediately followed by a return you could just get the error handler when you need it instead of pre-emptively here. I would probably rework `asyncErrHandlerFrom` into `maybeReportError(ctx context.Context, err error)` which does the context value unpacking and calls any discovered function. That also removes the need for `noopErrHandler` since `maybeReportError` would just return if no error handler was found in the context.

v2/blockstore/readonly.go
mvdan: this is gofmt; it will generally attempt to do vertical alignment if you group things together without empty lines. in this case, the grouping+alignment isn't gaining you much.

v2/blockstore/readonly.go
iand: It used to be recommended that context value keys be strings to avoid allocations. I'm not sure what the status is but this issue seems to suggest that strings are still preferred: https://github.com/golang/go/issues/33742

masih: Thank you @iand for pointing this out; will apply 🍻

mvdan (approved):
LGTM - just need to fix the conflicts I think

masih:
Thank you both for the speedy reviews 🚀

Issue #171, opened 2021-07-16
add opt-in support for treating null padding as EOF in v0 (CARv1)


mvdan:
https://github.com/ipld/go-car/issues/142 is about the v2 module, so this issue is about also propagating the feature to the v0 module, as per https://github.com/ipld/go-car/issues/140#issuecomment-874522993.

masih:
Note, util functions in v1 are exported and used it seems:
* https://grep.app/search?q=go-car/util&filter[lang][0]=Go
* https://grep.app/search?q=go-car/util

mvdan:
This is not urgent, as the only need for this is Filecoin, and they can't upgrade to go-car v0 master until the ipld-prime upgrade via graphsync takes place.

rvagg:
closing this out as we've largely migrated lotus away from direct CARv1 usage and this isn't needed there anymore (as far as I recall anyway)

someone's welcome to reopen if they believe this is important but we'd like to keep CARv1 stable/in-stasis as much as possible now

PR #191, opened 2021-07-23
Add zero-length sections as EOF option to internal CARv1 reader

### LINKED ISSUE #190: Propagate options to internal CARv1 in CARv2
Propagate options to the internal CARv1 used in CARv2. Right now no options are propagated. There is a single option we are looking to propagate here which is `ZeroLengthSectionAsEOF`

masih:
The CARv2 implementation uses an internal fork of CARv1 due to upstream
dependency issues captured in #104. Propagate the options set in CARv2
APIs for treatment of zero-lenth sections onto internal packages so that
APIs using the internal CARv1 reader behave consistently. Use a `bool`
for setting the option, since it is the only option needed in CARv1.

Update tests to reflect changes.

Add additional tests to internal CARv1 package and ReadOnly blockstore
that assert option is propagated.

Fixes #190

### DIFF
--- a/v2/blockstore/readonly.go
+++ b/v2/blockstore/readonly.go
@@ -162,7 +162,7 @@ func OpenReadOnly(path string, opts ...carv2.ReadOption) (*ReadOnly, error) {
 }
 
 func (b *ReadOnly) readBlock(idx int64) (cid.Cid, []byte, error) {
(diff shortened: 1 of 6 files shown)


mvdan (approved):

mvdan:
You can take https://github.com/ipld/go-car/issues/171 too if you want, since it would be extremely similar. That's not urgent though, so you could do it when you're back from holiday.

PR #192, opened 2021-07-23
Implement reader block iterator over CARv1 or CARv2

### LINKED ISSUE #189: Implement block iterator in CARv2 reader that accepts both v1 and v2
CARv1 reader provides an interface for iterating through sections [here](https://github.com/ipld/go-car/blob/master/car.go#L147). Ideally, we want a similar API for CARv2 reader that does the right thing given v1 or v2 payload

Cc @willscott

masih:
Implement the mechanism to iterate over blocks in CARv1 or CARv2 via the
`carv2.Reader` API.

Implement tests both for the iterator and the rest of the reader API.
Add test files with corrupt sections etc.

Fixes #189


v2/reader.go
```diff
@@ -35,32 +39,39 @@ func OpenReader(path string, opts ...ReadOption) (*Reader, error) {
...
 }
 
-// NewReader constructs a new reader that reads CARv2 from the given r.
-// Upon instantiation, the reader inspects the payload by reading the pragma and will return
-// an error if the pragma does not represent a CARv2.
+// NewReader constructs a new reader that reads either CARv1 or CARv2 from the given r.
+// Upon instantiation, the reader inspects the payload and provides appropriate read operations
+// for both CARv1 and CARv2.
+//
+// Note, Any other version other than 1 or 2 will result in an error. The caller may use
```
mvdan, line 46:
```suggestion
// Note that any version other than 1 or 2 will result in an error. The caller may use
```

v2/reader_test.go
```diff
@@ -0,0 +1,246 @@
...
+			"CarV2WithValidHeaderButCorruptSectionAndIndexIsTwo",
+			"testdata/sample-v2-corrupt-data-and-index.car",
+			2,
+			false,
+		},
+		{
+			"CarFileWithCorruptPragmaIsError",
+			"testdata/sample-corrupt-pragma.car",
+			0,
+			true,
```
mvdan, line 73: please key all these fields :) as a bonus, you can then get rid of the lines where the values are zero (0, false, etc)

v2/reader_test.go
```diff
@@ -0,0 +1,246 @@
...
+		},
+		{
+			"AnotherCarV1WithZeroLenSectionWithoutOption",
+			"testdata/sample-v1-with-zero-len-section2.car",
+			false,
+		},
+		{
+			"CorruptCarV1",
+			"testdata/sample-v1-tailing-corrupt-section.car",
+			false,
```
mvdan, line 137: ditto

v2/reader.go
```diff
@@ -13,10 +14,13 @@ import (
...
-	Header Header
-	r      io.ReaderAt
-	roots  []cid.Cid
-	closer io.Closer
+	Header      Header
+	Version     uint64
+	r           io.ReaderAt
+	roots       []cid.Cid
+	ropts       ReadOptions
+	carV1Reader *carv1.CarReader
```
mvdan, line 22: document that this is only used for Next? for example, it's not obvious that it can't be reused when the user asks for a carv1 reader - because then the offsets would be shared too.

v2/reader.go
```diff
@@ -2,6 +2,7 @@ package car
 
 import (
 	"fmt"
+	blocks "github.com/ipfs/go-block-format"
```
mvdan, line 5: nit: don't put third party imports in the std group. I think gofumpt fixes this.

mvdan (approved):

v2/reader.go
masih: Will add extra comments here thanks for pointing this out. I have already added docs to the [`Next`](https://github.com/ipld/go-car/blob/78dd9e9e920e21b62cdd5844c4fdc10478911653/v2/reader.go#L133) function pointing this out.

PR #193, opened 2021-07-27
add the first read-only benchmarks

mvdan:
(see commit message)

### DIFF
--- a/v2/bench_test.go
+++ b/v2/bench_test.go
@@ -1,0 +1,48 @@
+package car_test
+
+import (
+	"io"
+	"os"
+	"testing"
+
+	carv2 "github.com/ipld/go-car/v2"
+)
+
+// Open a reader, get the roots, and iterate over all blocks.
+// Essentially looking at the contents of any CARv1 or CARv2 file.
+// Note that this also uses ReadVersion underneath.
+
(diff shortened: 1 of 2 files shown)


mvdan:
Let me know if either of you think we should have more benchmarks. The only other I have in mind to do later is `blockstore.BenchmarkOpenReadWrite`, with its insertion index, Put deduplication, and "finalize" logic.

I generally prefer fewer benchmarks that cover more ground in terms of exposed API. 3 benchmarks that cover 90% of the code are more usable than 30 benchmarks that cover 99% of the code, I think - it's always easy to re-run 3 benchmarks after a big refactor and get an idea of how the performance changed, but running tens of benchmarks can get really slow and painful. So I'm explicitly not benchmarking edge cases and custom options right now, unless we have a good reason to :)

mvdan:
There's also the question of whether we should split the benchmarks by input size, such as OpenReadOnly on a 10KiB CARv1, then a 1MiB, then a 100MiB. That depends on what kinds of CAR files we expect downstream users to be encountering when the go-car library becomes a bottleneck.

willscott (approved):
I think these are fine as a baseline view of file and index accesses. we'll see other issues more usefully in datastore benchmarks a level above this library.

PR #194, opened 2021-07-30
update package godocs and root level README for v2

mvdan:
(see commit message)


README.md
```diff
@@ -4,17 +4,24 @@ go-car (go!)
 [![](https://img.shields.io/badge/made%20by-Protocol%20Labs-blue.svg?style=flat-square)](https://protocol.ai)
 [![](https://img.shields.io/badge/project-ipld-orange.svg?style=flat-square)](https://github.com/ipld/ipld)
 [![](https://img.shields.io/badge/freenode-%23ipld-orange.svg?style=flat-square)](https://webchat.freenode.net/?channels=%23ipld)
```
willscott, line 6: update to current ipld matrix channel?

willscott (approved):

willscott:
a short code example in the primary readme wouldn't hurt either

README.md
mvdan: Do you have a badge handy for that?

mvdan:
> a short code example in the primary readme wouldn't hurt either

I thought about it but I'm conflicted. We have examples in the godoc, which are great because they are runnable, so we'll notice immediately if they break. A README example doesn't have that, and isn't particularly visible from godoc either.

Maybe we instead explicitly say that godoc has examples?

willscott:
if you just want a pointer, that's fine with me.
I do approach a lot of our ipfs/ipld repos by starting at the readme though, and the current state of it doesn't shed as much light as we might like on recommended path for implementation.

masih:
A follow-on PR coming up adding links to examples etc.

PR #195, opened 2021-08-02
blockstore: use errors when API contracts are broken

mvdan:
(see commit message)


mvdan:
Err, I forgot to run the tests.

v2/blockstore/readwrite.go
```diff
@@ -284,22 +286,22 @@ func (b *ReadWrite) unfinalize() error {
 	return err
 }
 
-func (b *ReadWrite) panicIfFinalized() {
-	if b.header.DataSize != 0 {
-		panic("must not use a read-write blockstore after finalizing")
-	}
+func (b *ReadWrite) finalized() bool {
+	return b.header.DataSize != 0
```
dirkmc, line 290: Is it safe to access this value outside of a lock?

mvdan: Finalize is the only one writing to it. So it's only unsafe if one calls Finalize twice concurrently. Calling Finalize twice already errors, and the method is not documented as safe for concurrent use.

dirkmc (approved):

masih (approved):

PR #198, opened 2021-08-03
Implement version agnostic streaming CAR block iterator

masih:
Implement an iterator over CAR blocks that accepts both v1 and v2. The
reader only requires io.Reader interface to operate and exposes the
version and roots of the underlaying CAR file as fields.

Remove `Next` from CARv2 reader now that the iteration functionality is
provided using `BlockReader`.

Update the benchmarks to reflect the changes and add example usage for
`NewBlockReader`.

Add tests that assert `BlockReader` behaviour in erroneous and
successful cases for both underlying CARv1 and CARv2 payload.

### DIFF
--- a/v2/bench_test.go
+++ b/v2/bench_test.go
@@ -8,10 +8,9 @@ import (
 	carv2 "github.com/ipld/go-car/v2"
 )
 
-// Open a reader, get the roots, and iterate over all blocks.
-// Essentially looking at the contents of any CARv1 or CARv2 file.
-// Note that this also uses ReadVersion underneath.
-
+// BenchmarkReadBlocks instantiates a BlockReader, and iterates over all blocks.
+// It essentially looks at the contents of any CARv1 or CARv2 file.
+// Note that this also uses internal carv1.ReadHeader underneath.
 func BenchmarkReadBlocks(b *testing.B) {
 	path := "testdata/sample-wrapped-v2.car"
 
@@ -24,16 +23,16 @@ func BenchmarkReadBlocks(b *testing.B) {
 
 	b.RunParallel(func(pb *testing.PB) {
 		for pb.Next() {
-			cr, err := carv2.OpenReader(path)
+			r, err := os.Open("testdata/sample-wrapped-v2.car")
 			if err != nil {
 				b.Fatal(err)
 			}
-			_, err = cr.Roots()
+			br, err := carv2.NewBlockReader(r)
 			if err != nil {
 				b.Fatal(err)
 			}
 			for {
-				_, err := cr.Next()
+				_, err := br.Next()
 				if err == io.EOF {
 					break
 				}
@@ -42,7 +41,9 @@ func BenchmarkReadBlocks(b *testing.B) {
 				}
 			}
 
-			cr.Close()
+			if err := r.Close(); err != nil {
+				b.Fatal(err)
+			}
 		}
 	})
 }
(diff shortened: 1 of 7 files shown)


v2/bench_test.go
```diff
@@ -42,7 +41,9 @@ func BenchmarkReadBlocks(b *testing.B) {
 				}
 			}
 
-			cr.Close()
+			if err := r.Close(); err != nil {
+				b.Fatal(err)
+			}
```
mvdan, line 46: FYI, no need to check for an error, since we just read

v2/block_reader.go
```diff
@@ -0,0 +1,142 @@
...
+		}
+
+		// Skip to the beginning of inner CARv1 data payload.
+		// Note, at this point the pragma and CARv1 header have been read.
+		// An io.ReadSeeker is opportunistically constructed from the given io.Reader r.
+		// The constructor does not take an initial offset, so we use Seek in io.SeekCurrent to
+		// fast forward to the beginning of data payload by subtracting pragma and header size from
+		// dataOffset.
+		rs := internalio.ToByteReadSeeker(r)
+		if _, err := rs.Seek(dataOffset-PragmaSize-HeaderSize, io.SeekCurrent); err != nil {
```
mvdan, line 91: I think the math here might be easier if you use SeekStart?

v2/example_test.go
```diff
@@ -68,3 +69,61 @@ func ExampleWrapV1File() {
...
+	// 	bafy2bzacedjwekyjresrwjqj4n2r5bnuuu3klncgjo2r3slsp6wgqb37sz4ck
+	// File: testdata/sample-wrapped-v2.car
+	// Version: 2
+	// Roots: [bafy2bzaced4ueelaegfs5fqu4tzsh6ywbbpfk3cxppupmxfdhbpbhzawfw5oy]
+	// First 5 block CIDs:
+	// 	bafy2bzaced4ueelaegfs5fqu4tzsh6ywbbpfk3cxppupmxfdhbpbhzawfw5oy
+	// 	bafy2bzaceaycv7jhaegckatnncu5yugzkrnzeqsppzegufr35lroxxnsnpspu
+	// 	bafy2bzaceb62wdepofqu34afqhbcn4a7jziwblt2ih5hhqqm6zitd3qpzhdp4
+	// 	bafy2bzaceb3utcspm5jqcdqpih3ztbaztv7yunzkiyfq7up7xmokpxemwgu5u
+	// 	bafy2bzacedjwekyjresrwjqj4n2r5bnuuu3klncgjo2r3slsp6wgqb37sz4ck
```
mvdan, line 128: nice output :) I had forgotten tabs works fine with examples

mvdan (approved):
:tada:

v2/block_reader.go
masih: I tried to explain why `SeekCurrent` is used in the comments above. TLDR; using `SeekStart` _could_ advance too far if `r` does not implement `io.ReadSeeker`. The opportunistic conversion from `io.Reader` to `io.ReadSeeker` does not take an initial offset, because we don't know if we can take an initial offset.

v2/bench_test.go
masih: Trying to be tidy, since close _can_ return an error. might be an overkill here agreed.

v2/block_reader.go
mvdan: Ah, I see. Makes sense to me.

PR #202, opened 2021-08-05
v2: stop using a symlink for LICENSE.md

mvdan:
(see commit message)


masih (approved):
Would removal of LICENSE from root break the v0 godoc ?

mvdan:
It wouldn't break the last v0 release as that's fixed, but it would make any future v0 releases and master non-OSS, which is also a problem.

And it would make GitHub's heuristics think this repo is non-OSS, probably :)

masih:
> It wouldn't break the last v0 release as that's fixed, but it would make any future v0 releases and master non-OSS, which is also a problem.
> 
> And it would make GitHub's heuristics think this repo is non-OSS, probably :)

I am curious then why are we removing the license from v0 at all?

mvdan:
Oh my. Definitely unintended. Second facepalm of the day.

Issue #205, opened 2021-08-09
ReadOnly blockstore should safely reject reads after Close() was called


raulk:
Right now, it will lead to a panic, like the one seen in https://github.com/filecoin-project/lotus/issues/7005. If opened with `OpenReadOnly()`, `Close()` will close the backing mmap and the next read will cause a panic.

PR #206, opened 2021-08-10
v2/blockstore: add ReadWrite.Discard

mvdan:
(see commit message)


v2/blockstore/readonly.go
```diff
@@ -331,6 +347,10 @@ func (b *ReadOnly) AllKeysChan(ctx context.Context) (<-chan cid.Cid, error) {
 	// We release the lock when the channel-sending goroutine stops.
 	b.mu.RLock()
 
+	if b.closed {
+		return nil, errClosed
+	}
```
masih, line 352: What would happen if blockstore is closed while iterating through keys? Do we want to return consistent error there too ?

v2/blockstore/readonly_test.go
```diff
@@ -235,3 +235,24 @@ func newV1Reader(r io.Reader, zeroLenSectionAsEOF bool) (*carv1.CarReader, error
...
+func TestReadOnlyErrorAfterClose(t *testing.T) {
+	bs, err := OpenReadOnly("../testdata/sample-v1.car")
+	require.NoError(t, err)
+
+	roots, err := bs.Roots()
+	require.NoError(t, err)
+	_, err = bs.Get(roots[0])
+	require.NoError(t, err)
+	_, err = bs.GetSize(roots[0])
+	require.NoError(t, err)
```
masih, line 248: no error assertion is probably an overkill since we have tests already that assert readonly with that file behaves.

v2/blockstore/readonly_test.go
```diff
@@ -235,3 +235,24 @@ func newV1Reader(r io.Reader, zeroLenSectionAsEOF bool) (*carv1.CarReader, error
...
+
+	bs.Close()
+
+	_, err = bs.Roots()
+	require.Error(t, err)
+	_, err = bs.Get(roots[0])
+	require.Error(t, err)
+	_, err = bs.GetSize(roots[0])
+	require.Error(t, err)
+}
```
masih, line 258: Would it make sense to add `AllKeysChan` while at it? 

A separate test-case for close with `AllKeysChan` mid iteration would also be good.

v2/blockstore/readwrite_test.go
```diff
@@ -405,7 +405,7 @@ func TestBlockstoreResumption(t *testing.T) {
 			require.Equal(t, wantBlockCountSoFar, gotBlockCountSoFar)
 		}
 	}
-	require.NoError(t, blockstore.CloseReadWrite(subject))
+	subject.Discard()
```
masih, line 408: Would be nice to have explicit test for discard now that it is a public API. 

Also separate tests for behaviour or RW blockstore after discard is called.

masih (approved):
One question about what would happen if backing is closed mid `AllKeysChan` iteration otherwise no blockers. Thank you for rolling this out so quickly

v2/blockstore/readonly.go
mvdan: We use mutexes, so a Close/Discard call will block if an AllKeysChan operation is happening. The docs already warn about this.

masih:
Fixes #205

v2/blockstore/readonly_test.go
mvdan: maybe overkill, I just wanted to show the mirror image of "does not error" with "always errors" before and after closing.

mvdan:
> Fixes #205

See the commit message :wink:

v2/blockstore/readonly.go
masih: OK that makes sense; I see `closeWithoutMutex` grabs the lock both in RW and RO; all good 👍

masih:
Odd.. GH did not link up the issue not even after my comment. I think this needs to be in PR description to work maybe? 

I ended up manually linking up to the issue.

mvdan:
Yeah the linking is kinda annoying. But the issue is closed when this PR is merged anyway, the mention in the commit message is still picked up that way.

v2/blockstore/readonly_test.go
mvdan: > Would it make sense to add `AllKeysChan` while at it?

Alright, I've added all read-only methods.

> A separate test-case for close with `AllKeysChan` mid iteration would also be good.

I agree this would be nice, but we already lacked a test before, and I want to merge this sooner than later for ignite. I'll just add a TODO.

v2/blockstore/readwrite_test.go
mvdan: done

mvdan:
@masih applied all your suggestions, including tests and the AllKeysChan mutex fix. Do you want to take one last look and merge on green?

masih (approved):

PR #208, opened 2021-08-10
Implement utility to extract CARv1 from a CARv2

### LINKED ISSUE #207: efficient buffer-less ExtractV1File method
CARv2 embeds a CARv1 fully, including its header. It should be possible thus to unpack a CARv1 from a CARv2 file efficiently by `os.Truncate`ing at the end, and `io.Copy`ing through `*os.File` descriptors to allow Go to delegate to the kernel instead of performing the copy in user land.

Two use cases are requested:
1. Extract inner CARv1 into a new file.
2. Overwrite existing CARv2 with the CARv1 contents (potentially by opening two `*os.File` file descriptors against the same file, seeking in the source descriptor to the right offset, and performing `io.Copy`?).

masih:
Implement `ExtractV1File` where the function takes path to a CARv2 file
and efficiently extracts its inner CARv1 payload. Note, the
implementation only supports CARv2 as input and returns a dedicated
error if the supplied input is already in CARv1 format.

Implement benchmarks comparing extraction using `Reader` vs
`ExtractV1File`.

Implement tests that assert in-place extraction as well as invalid input
and both v1/v2 input

Fixes #207

### DIFF
--- a/v2/bench_test.go
+++ b/v2/bench_test.go
@@ -14,8 +14,6 @@ import (
 	carv2 "github.com/ipld/go-car/v2"
 )
 
-var rng = rand.New(rand.NewSource(1413))
-
 // BenchmarkReadBlocks instantiates a BlockReader, and iterates over all blocks.
 // It essentially looks at the contents of any CARv1 or CARv2 file.
 // Note that this also uses internal carv1.ReadHeader underneath.
@@ -59,7 +57,7 @@ func BenchmarkReadBlocks(b *testing.B) {
 // BenchmarkExtractV1File extracts inner CARv1 payload from a sample CARv2 file using ExtractV1File.
 func BenchmarkExtractV1File(b *testing.B) {
 	path := filepath.Join(b.TempDir(), "bench-large-v2.car")
-	generateRandomCarV2File(b, path, 10*1024*1024) // 10 MiB
+	generateRandomCarV2File(b, path, 10<<20) // 10 MiB
 	defer os.Remove(path)
 
 	info, err := os.Stat(path)
@@ -87,7 +85,7 @@ func BenchmarkExtractV1File(b *testing.B) {
 // BenchmarkExtractV1File.
 func BenchmarkExtractV1UsingReader(b *testing.B) {
 	path := filepath.Join(b.TempDir(), "bench-large-v2.car")
-	generateRandomCarV2File(b, path, 10*1024*1024) // 10 MiB
+	generateRandomCarV2File(b, path, 10<<20) // 10 MiB
 	defer os.Remove(path)
 
 	info, err := os.Stat(path)
@@ -121,6 +119,8 @@ func BenchmarkExtractV1UsingReader(b *testing.B) {
 }
 
 func generateRandomCarV2File(b *testing.B, path string, minTotalBlockSize int) {
+	// Use fixed RNG for determinism across benchmarks.
+	rng := rand.New(rand.NewSource(1413))
 	bs, err := blockstore.OpenReadWrite(path, []cid.Cid{})
 	defer func() {
 		if err := bs.Finalize(); err != nil {
@@ -130,7 +130,7 @@ func generateRandomCarV2File(b *testing.B, path string, minTotalBlockSize int) {
 	if err != nil {
 		b.Fatal(err)
 	}
-	buf := make([]byte, 1024)
+	buf := make([]byte, 32<<10) // 32 KiB
 	var totalBlockSize int
 	for totalBlockSize < minTotalBlockSize {
 		size, err := rng.Read(buf)
--- a/v2/bench_test.go
+++ b/v2/bench_test.go
@@ -2,11 +2,19 @@ package car_test
 
 import (
 	"io"
+	"math/rand"
 	"os"
+	"path/filepath"
 	"testing"
+
+	"github.com/ipfs/go-cid"
+	"github.com/ipfs/go-merkledag"
+	"github.com/ipld/go-car/v2/blockstore"
 
 	carv2 "github.com/ipld/go-car/v2"
 )
+
+var rng = rand.New(rand.NewSource(1413))
 
 // BenchmarkReadBlocks instantiates a BlockReader, and iterates over all blocks.
 // It essentially looks at the contents of any CARv1 or CARv2 file.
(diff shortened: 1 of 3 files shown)


v2/writer.go
```diff
@@ -79,6 +80,99 @@ func WrapV1(src io.ReadSeeker, dst io.Writer) error {
...
+		// Copy all of src into dst.
+		if _, err := io.Copy(dst, src); err != nil {
+			return err
+		}
+	case 2:
+		// Read CARv2 header to locate data payload.
+		v2h := &Header{}
+		if _, err := v2h.ReadFrom(src); err != nil {
+			return err
+		}
```
raulk, line 133: This won't escape to heap, but this is more idiomatic anyway:

```suggestion
		var v2h Header
		if _, err := v2h.ReadFrom(&src); err != nil {
			return err
		}
```

v2/writer.go
```diff
@@ -79,6 +80,99 @@ func WrapV1(src io.ReadSeeker, dst io.Writer) error {
...
+			return fmt.Errorf("invalid data payload size: %v", dataSize)
+		}
+
+		// Seek to the point where the data payload starts
+		if _, err := src.Seek(dataOffset, io.SeekStart); err != nil {
+			return err
+		}
+
+		// Copy data payload over, expecting to write exactly the right number of bytes
+		written, err := io.CopyN(dst, src, dataSize)
```
raulk, line 152: Are we sure that this will delegate to kernel? If so, can we add a comment here explaining what we expect to happen?

v2/writer.go
```diff
@@ -79,6 +80,99 @@ func WrapV1(src io.ReadSeeker, dst io.Writer) error {
...
+		if err != nil {
+			return err
+		}
+		if written != dataSize {
+			return fmt.Errorf("expected to write exactly %v but wrote %v", dataSize, written)
+		}
+		// Truncate to the expected size to assure the resulting file is a correct CARv1.
+		if err = dst.Truncate(written); err != nil {
+			return err
+		}
```
raulk, line 162: Is this necessary if we're using `CopyN`?

v2/writer.go
```diff
@@ -79,6 +80,99 @@ func WrapV1(src io.ReadSeeker, dst io.Writer) error {
...
+	default:
+		return fmt.Errorf("invalid source version: %v", version)
+	}
+
+	// Check the close error, since we're writing to dst.
+	// Note that we also do a "defer dst.Close()" above,
```
raulk, line 168: Is it safe to double close?

v2/writer.go
```diff
@@ -79,6 +80,99 @@ func WrapV1(src io.ReadSeeker, dst io.Writer) error {
...
+	// to make sure that the earlier error returns don't leak the file.
+	if err := dst.Close(); err != nil {
+		return err
+	}
```
raulk, line 172: Do we want to return the error on close? We may want to just absorb it? (or log it, but this library shouldn't log I think)

v2/writer.go
masih: Comments added. thank you for pointing this out.

 Re "Are we sure":
I have implemented two benchmarks: one uses `ExtractV1File`, and the other user `Reader` API. The expectation is that the former is more efficient. The point is to measure this.

v2/writer.go
masih: I lean towards a yes but happy to be convinced otherwise. Generally following advice from posts like [this](https://www.joeshaw.org/dont-defer-close-on-writable-files/). There are other places in the repo where close errors are indeed returned. 

Deferring any changes to a separate PR until we decide how to approach this consistently.

mvdan:
Just so we're all on the same page, doing an efficient copy via the OS is best-effort. For example, on Linux, we'll fall back to a regular old user-space copy if the kernel is too old: https://cs.opensource.google/go/go/+/refs/tags/go1.16.7:src/internal/poll/copy_file_range_linux.go;drc=refs%2Ftags%2Fgo1.16.7;l=54

With @masih we agreed to briefly document the best-effort efficiency in the API docs.

v2/bench_test.go
```diff
@@ -2,12 +2,20 @@ package car_test
...
 	"testing"
 
+	"github.com/ipfs/go-cid"
+	"github.com/ipfs/go-merkledag"
+	"github.com/ipld/go-car/v2/blockstore"
+
 	carv2 "github.com/ipld/go-car/v2"
 )
 
+var rng = rand.New(rand.NewSource(1413))
```
mvdan, line 17: nit: move this inside generateRandomCarV2File, because then each call to that function will be deterministic and produce the same file. otherwise running the benchmarks in a different order (or only running one of them) can cause different behavior and resulting numbers.

plus, fewer globals.

v2/bench_test.go
```diff
@@ -47,3 +55,93 @@ func BenchmarkReadBlocks(b *testing.B) {
 		}
 	})
 }
+
+// BenchmarkExtractV1File extracts inner CARv1 payload from a sample CARv2 file using ExtractV1File.
+func BenchmarkExtractV1File(b *testing.B) {
+	path := filepath.Join(b.TempDir(), "bench-large-v2.car")
+	generateRandomCarV2File(b, path, 10*1024*1024) // 10 MiB
```
mvdan, line 62: neat trick if you weren't aware: `10 << 20 // 10 MiB`.

`<< 10` for KiB, `<< 20` for MiB, `<< 30` for GiB, and so on. Easier to type and read, if you ask me, as the "metric scale" is a number that increases 1-2-3-etc.

also pretty intuitive once you remember that `1 << 10 == 1024`

v2/writer.go
```diff
@@ -79,6 +84,109 @@ func WrapV1(src io.ReadSeeker, dst io.Writer) error {
...
+	// Detect CAR version.
+	version, err := ReadVersion(src)
+	if err != nil {
+		return err
+	}
+	if version == 1 {
+		return ErrAlreadyV1
+	}
+	if version != 2 {
+		return fmt.Errorf("invalid source version: %v", version)
```
mvdan, line 120: maybe say "source version must be 2" to make it clearer? invalid makes me think a version is truly bad, like -3.

v2/writer.go
```diff
@@ -79,6 +84,109 @@ func WrapV1(src io.ReadSeeker, dst io.Writer) error {
...
+
+	// TODO consider extracting this into Header.Validate since it is also implemented in BlockReader.
+	// Validate header
+	dataOffset := int64(v2h.DataOffset)
+	if dataOffset < PragmaSize+HeaderSize {
+		return fmt.Errorf("invalid data payload offset: %v", dataOffset)
+	}
+	dataSize := int64(v2h.DataSize)
+	if dataSize <= 0 {
+		return fmt.Errorf("invalid data payload size: %v", dataSize)
```
mvdan, line 137: nit: you know these are integers, you can use `%d` instead of `%v`

v2/writer.go
```diff
@@ -79,6 +84,109 @@ func WrapV1(src io.ReadSeeker, dst io.Writer) error {
...
+	if _, err := src.Seek(dataOffset, io.SeekStart); err != nil {
+		return err
+	}
+
+	// Open destination as late as possible to minimise unintended file creation in case an error
+	// occurs earlier.
+	// Note, we explicitly do not use os.O_TRUNC here so that we can support in-place extraction.
+	// Otherwise, truncation of an existing file will wipe the data we would be reading from if
+	// source and destination paths are the same.
+	// Later, we do truncate the file to the right size to assert there are no tailing extra bytes.
```
mvdan, line 150: :ok_hand:

v2/bench_test.go
```diff
@@ -47,3 +55,93 @@ func BenchmarkReadBlocks(b *testing.B) {
...
+	bs, err := blockstore.OpenReadWrite(path, []cid.Cid{})
+	defer func() {
+		if err := bs.Finalize(); err != nil {
+			b.Fatal(err)
+		}
+	}()
+	if err != nil {
+		b.Fatal(err)
+	}
+	buf := make([]byte, 1024)
```
mvdan, line 133: nit: I think it would be slightly more realistic if the blocks are a bit larger, like 32KiB. The block sizes could really be anything, but in general one can guess that the average block size will be in the tens-to-hundreds of kilobytes.

for example, I believe IPFS starts chunking files into blocks if they are over 256KiB.

mvdan (approved):
LGTM. I'll be back with benchmark numbers.

mvdan:
For future record, the ~15% improvement on Linux with tmpfs is not very realistic. When using the SSD in the same laptop, I now get:

```
name         old time/op    new time/op    delta
ExtractV1-8    8.83ms ±29%    1.43ms ± 2%   -83.83%  (p=0.000 n=8+8)

name         old speed      new speed      delta
ExtractV1-8  1.24GB/s ±33%  7.36GB/s ± 2%  +494.34%  (p=0.000 n=8+8)

name         old alloc/op   new alloc/op   delta
ExtractV1-8    34.0kB ± 1%     1.0kB ± 0%   -97.20%  (p=0.000 n=8+8)

name         old allocs/op  new allocs/op  delta
ExtractV1-8      26.0 ± 0%      23.0 ± 0%   -11.54%  (p=0.000 n=8+8)
```

So the improvement seems to be anywhere from 10 to 80%, depending on your disk.

Issue #214, opened 2021-08-27
New Index format that contains the hashing algorithm information


masih:
The current `IndexSorted` format contains the digest of the CIDs within in a CAR and offset to where the data is located. This format does not contain enough information to reconstruct list of multihashes in the CAR from index alone, because it does not store the hash function used to generate the digests.

Data providers to [indexer nodes](https://github.com/filecoin-project/storetheindex) should ideally be able to use the CAR index alone to supply the list of CIDs in a CAR without having to scan the entire CAR file. When data providers run as part of FileCoin miners, there will be indeed cases where the only accessible information about a CAR is its detached index. 

Since for indexing purposes, the codec information in CID is ignored (see https://github.com/ipfs/go-ipfs/issues/6815) the CAR index should at least expose the hash function used to generate the digests within the index. This way the data providers are able to re-construct the list of CAR multihashes.

Issue #215, opened 2021-08-30
`IndexSorted` should not load CIDs with multihash code of `IDENTITY`


masih:
It is possible for a CAR file to contain CIDs with multihash `IDENTITY`. Such CIDs could significantly bloat the index size and cause OOM error. As stated in the [CARv2 spec](https://ipld.io/specs/transport/car/carv2/#index-format) Index should disallow them and safely skip over them when indexing a CAR file.

`IndexSorted` currently loads all given CIDs. Skip over CIDs with multihash `IDENTITY`.

rvagg:
Indeedy, we even included this in the [spec](https://ipld.io/specs/transport/car/carv2/#index-format) and made it bold:

> Indexes **should not include identity hash CIDs**. It is assumed that any use of a CARv2 as a blockstore will return identity CID data immediately by extracting it from the CID, therefore there should be no need to provide indexing for such entries.

masih:
The new index type does this already; I'll get this sorted in the `IndexSorted` type shortly

PR #217, opened 2021-09-01
Implement new index type that also includes mutltihash code

masih:
Implement a new CARv2 index that contains enough information to
reconstruct the multihashes of the data payload, since `CarIndexSorted`
only includes multihash digests. The new index builds on top of the 
existing `IndexSorted` by adding an additional layer of grouping the 
multi-width indices by their multihash code.

Note, this index intentionally ignores
any given record with `multihash.IDENTITY` CID hash.

Add a test that asserts offsets for the same CID across sorted index and
new multihash sorted index are consistent.

Add tests that assert marshal unmarshalling of the new index type is as
expected, and it does not load records with `multihash.IDENTITY` digest.

Relates to:
- https://github.com/multiformats/multicodec/pull/227

Fixes https://github.com/ipld/go-car/issues/214

### DIFF
--- a/v2/index/mhindexsorted.go
+++ b/v2/index/mhindexsorted.go
@@ -1,0 +1,158 @@
+package index
+
+import (
+	"encoding/binary"
+	"io"
+	"sort"
+
+	"github.com/ipfs/go-cid"
+	"github.com/multiformats/go-multicodec"
+	"github.com/multiformats/go-multihash"
+)
+
+type (
+	// multihashIndexSorted maps multihash code (i.e. hashing algorithm) to multiWidthCodedIndex.
+	// This index ignores any Record with multihash.IDENTITY.
+	multihashIndexSorted map[uint64]*multiWidthCodedIndex
+	// multiWidthCodedIndex stores multihash code for each multiWidthIndex.
+	multiWidthCodedIndex struct {
+		multiWidthIndex
+		code uint64
+	}
+)
+
+func newMultiWidthCodedIndex() *multiWidthCodedIndex {
+	return &multiWidthCodedIndex{
+		multiWidthIndex: make(multiWidthIndex),
+	}
+}
+
+func (m *multiWidthCodedIndex) Marshal(w io.Writer) error {
+	if err := binary.Write(w, binary.LittleEndian, m.code); err != nil {
+		return err
+	}
+	return m.multiWidthIndex.Marshal(w)
+}
+
+func (m *multiWidthCodedIndex) Unmarshal(r io.Reader) error {
+	if err := binary.Read(r, binary.LittleEndian, &m.code); err != nil {
+		return err
+	}
+	return m.multiWidthIndex.Unmarshal(r)
+}
+
+func (m *multihashIndexSorted) Codec() multicodec.Code {
+	return multicodec.CarMultihashIndexSorted
+}
+
+func (m *multihashIndexSorted) Marshal(w io.Writer) error {
+	if err := binary.Write(w, binary.LittleEndian, int32(len(*m))); err != nil {
+		return err
+	}
+	// The codes are unique, but ranging over a map isn't deterministic.
+	// As per the CARv2 spec, we must order buckets by digest length.
+	// TODO update CARv2 spec to reflect this for the new index type.
+	codes := m.sortedMultihashCodes()
+
+	for _, code := range codes {
+		mwci := (*m)[code]
+		if err := mwci.Marshal(w); err != nil {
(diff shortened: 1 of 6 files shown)


v2/index/indexmhsorted.go
```diff
@@ -0,0 +1,210 @@
...
+	"github.com/multiformats/go-multicodec"
+	"github.com/multiformats/go-multihash"
+)
+
+// TODO replace with propper multicodec once defined.
+const IndexMultihashSortedCodec = multicodec.Code(0x0401)
+
+type (
+	multiWidthCodedIndex  map[uint32]codedSingleWidthIndex
+	codedSingleWidthIndex map[uint64]singleWidthIndex
```
willscott, line 18: do we need two layers (bitwidth and then multihash code)? can we get away with only splitting the index by the multihash algorithm, and assume the width within a given multihash algorithm will always be the same?

willscott:
> @willscott This index implementation uses a nested map of `length` -> `multihash.Code` -> `singleWidthIndex`. I am not sure if we ever need to support hash functions that produce variable length digests. `IDENTITY` is an obvious one but that's a special case which we will be skipping over as captured in #215. If not the storage structure could be simplified.
> 
> I'd love to get your feedback whenever you get a chance. 🍻

Unless we can think of cases where we wouldn't be able to get away with the simpler case, lets see if we can get away with it?

ribasushi:
@masih @willscott one of these days when I carve some time to build what I was planning to build before joining PL, I am definitely going to use multihash-level truncated full-sha3-512 ( can go into the weeds why if interested ). So you won't be able to index my content e.g. ( whether this is a showstopper or not: 🤷 )

willscott:
that means the car will have to be read through block by block to calculate the list of full multi-hashes to provide for indexing, rather than being able to crib it off the index of the car. Eventually, it sounds like we should be able to store the computed announcement for indexer nodes if we do have to do the extra work of computation to generate the announcement. that'll probably happen after MVP (although @masih already wrote the walking code once 😅 )

v2/index/indexmhsorted.go
masih: Agreed; marking the PR as draft to get this change through. Thanks Will

v2/index_gen_test.go
```diff
@@ -153,3 +164,93 @@ func TestGenerateIndexFromFile(t *testing.T) {
...
+
+		cidLen, c, err := cid.CidFromReader(reader)
+		require.NoError(t, err)
+		records = append(records, index.Record{Cid: c, Offset: uint64(sectionOffset)})
+		remainingSectionLen := int64(sectionLen) - int64(cidLen)
+		sectionOffset, err = reader.Seek(remainingSectionLen, io.SeekCurrent)
+		require.NoError(t, err)
+	}
+
+	err = idx.Load(records)
```
willscott, line 252: extend to make sure you can serialize the index to disk and load it back again?

willscott:
@rvagg anything we should be aware of in putting together a PR for the additional index multicodec format over in multicodecs?

rvagg:
nope, should be pretty straightforward, just add a `0x0401` and it should get merged in short order

masih:
> nope, should be pretty straightforward, just add a `0x0401` and it should get merged in short order

thanks @rvagg, going to open a PR for it just now if we are happy with `0x0401` value

v2/index_gen_test.go
masih: Resolved [here](https://github.com/ipld/go-car/blob/c28358b0377ab1c4468197a4d0e5a0bb69cb7f0c/v2/index/mhindexsorted_test.go)

v2/index/mhindexsorted.go
```diff
@@ -0,0 +1,158 @@
...
+func (m *multihashIndexSorted) put(mwci *multiWidthCodedIndex) {
+	(*m)[mwci.code] = mwci
+}
+
+func (m *multihashIndexSorted) Load(records []Record) error {
+	// TODO optimize load by avoiding multihash decoding twice.
+	// This implementation decodes multihashes twice: once here to group by code, and once in the
+	// internals of multiWidthIndex to group by digest length. The code can be optimized by
+	// combining the grouping logic into one step where the multihash of a CID is only decoded once.
+	// The optimization would need refactoring of the IndexSorted compaction logic.
```
masih, line 101: Postponing the optimisation here to reduce the size of this PR

masih (commented):
Self-review

willscott (approved):

PR #218, opened 2021-09-02
Ignore records with `IDENTITY` CID in `IndexSorted`

### LINKED ISSUE #215: `IndexSorted` should not load CIDs with multihash code of `IDENTITY`

masih:
Skip `IDENTITY` CIDs in `IndexSorted`, as required by CARv2
specification.

Re-generate `sample-wrapped-v2.car` testdata, since its previous index
contained CIDs with `IDENTITY` multihash.

Gracefully handle CIDs with `IDENTITY` multihash code in the
`blockstore` API. Since `Get`, `GetSize` and `Has` interfaces rely on
the index, decode given keys directly to satisfy calls for such CIDs.

See:
- https://ipld.io/specs/transport/car/carv2/#index-format

Fixes #215


willscott:
I'm not totally opposed to this as a defensive mechanism.

in the past we've used the pattern of wrapping the blockstore in an idstore https://github.com/ipfs/go-ipfs-blockstore/blob/master/idstore.go when there is the potential for interaction with identity-hashed blocks.
I wonder if we're okay at this layer just panic-ing / being unhappy if they show up and expecting that wrapping layer to take care of them?

masih:
> I wonder if we're okay at this layer just panic-ing / being unhappy if they show up and expecting that wrapping layer to take care of them?

Yup that makes sense; will do 🍻

rvagg:
I'm a soft +1 on all our layers behaving like this, but it's a bit tricky since if we don't do it _completely_ then we set up a false expectation for users that they don't need to wrap or be careful with the layers they're interacting with. I don't think there's harm in doing it here, it's quite nice, especially if you just want to use a CARv2 as a raw blockstore, but the user expectation problem will be the thing that gets us, if anything.

masih:
> I'm a soft +1 on all our layers behaving like this, but it's a bit tricky since if we don't do it _completely_ then we set up a false expectation for users that they don't need to wrap or be careful with the layers they're interacting with. I don't think there's harm in doing it here, it's quite nice, especially if you just want to use a CARv2 as a raw blockstore, but the user expectation problem will be the thing that gets us, if anything.

Having implemented an explicit error when given `IDENTITY` CIDs in blockstore [here](https://github.com/ipld/go-car/pull/218/commits/ac3b3fa1eb716665c7d58e403ba472d2c5822cb4), I am leaning towards being more forgiving in the CARv2 blockstore APIs and "doing the right thing" when given such CIDs. 

This is because, in order to explicitly error out the blockstore needs to decode the CID hash, and right there we have all the information to do what [`IdStore`](https://github.com/ipfs/go-ipfs-blockstore/blob/master/idstore.go) does. Sure; it would be duplicating the logic of `IdStore` in CARv2 library, but not doing it and force wrapping means:
1) potential code changes for dependant work streams if CAR files interacted with via the blockstore API often contain `IDENTITY` CIDs (cc @aarshkshah1992)
2) the CIDs will get decoded twice: once by the wrapping `IdStore` and once by the blockstore API here to check that they are indeed _not_ `IDENTITY` CIDs.

@willscott @rvagg  @aarshkshah1992 What are your thoughts?

rvagg:
I think my weakly held position on this is that CARv2 is somewhat likely to be used as a stand-alone blockstore, for IPLD applications that exist outside of IPFS and want a simple immutable block store, so if the information is at hand already (i.e. it's cheap), I think it should transparently handle identity CIDs.

masih:
After much discussion, this PR now gracefully handles `IDENTITY` CIDs, doing what `IdStore` wrapper does without the need for wrapping.

Thanks @willscott @rvagg

v2/blockstore/readonly.go
```diff
@@ -320,6 +348,16 @@ func (b *ReadOnly) GetSize(key cid.Cid) (int, error) {
 	return fnSize, nil
 }
 
+func isIdentity(key cid.Cid) (ok bool, digest []byte, err error) {
```
mvdan, line 351: nit: usually the value would go before `ok bool`, and an error always goes last, so: `digest []byte, ok bool, err error`

v2/blockstore/readonly.go
```diff
@@ -320,6 +348,16 @@ func (b *ReadOnly) GetSize(key cid.Cid) (int, error) {
...
+	dmh, err := multihash.Decode(key.Hash())
+	if err != nil {
+		return
+	}
+	ok = dmh.Code == multihash.IDENTITY
+	digest = dmh.Digest
+	return
```
mvdan, line 358: best to not use naked returns if you can avoid it: https://github.com/golang/go/wiki/CodeReviewComments#naked-returns

I personally think they're only OK with the defer-error-check pattern, as you really need them in that case

mvdan (approved):
Should this behavior be mentioned in the blockstore package godoc? It seems pretty relevant at a high level.

v2/blockstore/doc.go
```diff
@@ -15,4 +15,14 @@
...
 // A user may resume reading/writing from files produced by an instance of ReadWrite blockstore. The
 // resumption is attempted automatically, if the path passed to OpenReadWrite exists.
+//
+// Note that the blockstore implementations in this package behave similarly to IPFS IdStore wrapper
+// when given CIDs with multihash.IDENTITY code.
+// More specifically, for CIDs with multhash.IDENTITY code:
+// * blockstore.Has will always return true
+// * blockstore.Get will always succeed, returning the multihash digest of the given CID
+// * blockstore.GetSize will always succeed, returning the multihash digest length of the given CID
+// * blockstore.Put and blockstore.PutMany will always succeed, without writing any bytes onto the backing file.
```
mvdan, line 25: how about "will always succeed without performing any operation"? because then it's also clear that it does not affect the index

v2/blockstore/readonly.go
```diff
@@ -8,6 +8,8 @@ import (
 	"io"
 	"sync"
 
+	"github.com/multiformats/go-multihash"
```
mvdan, line 11: put this with the third party import group?

mvdan (approved):

PR #219, opened 2021-09-07
Add carve utility for updating the index of a car{v1,v2} file

willscott:
run as
`carve index -c CarMultihashIndexSorted test.car`


carve/carve.go
```diff
@@ -0,0 +1,167 @@
...
+				Action: func(c *cli.Context) error {
+					r, err := carv2.OpenReader(c.Args().Get(0))
+					if err != nil {
+						return err
+					}
+					defer r.Close()
+
+					var idx index.Index
+					if c.String("codec") != "none" {
+						mc, ok := multicodec.Of(c.String("codec"))
```
mvdan, line 45: probably a good time to update go-car with https://github.com/multiformats/go-multicodec/pull/51:

```
var mc multicodec.Code
if err := mc.Set(c.String("codec")); err != nil {
    return err
}
```

carve/carve.go
```diff
@@ -0,0 +1,167 @@
...
+					if c.Args().Len() >= 2 {
+						outStream, err = os.Create(c.Args().Get(1))
+						if err != nil {
+							return err
+						}
+					}
+					defer outStream.Close()
+
+					v1r := r.DataReader()
+					v1Size, err := v1r.Seek(0, io.SeekEnd)
```
mvdan, line 65: you can just use `r.Header.DataSize` I think

carve/carve.go
```diff
@@ -0,0 +1,167 @@
...
+					records := make([]index.Record, 0)
+					var sectionOffset int64
+					if sectionOffset, err = v1r.Seek(0, io.SeekCurrent); err != nil {
+						return err
+					}
+
+					br := bufio.NewReader(v1r)
+					for {
+						// Read the section's length.
+						sectionLen, err := varint.ReadUvarint(br)
```
mvdan, line 113: can you not use carv2.NewBlockReader for this, instead of rolling your own reading of cids etc?

mvdan (commented):
Any particular reason you made another module? this could live in go-car/v2 as far as I can tell, like in v2/cmd/carve

carve/carve.go
```diff
@@ -0,0 +1,167 @@
...
+				Aliases: []string{"i"},
+				Usage:   "write out the car with an index",
+				Action: func(c *cli.Context) error {
+					r, err := carv2.OpenReader(c.Args().Get(0))
+					if err != nil {
+						return err
+					}
+					defer r.Close()
+
+					var idx index.Index
```
masih, line 43: We may have been given a CARv2 that has the desired index already; would it make sense to check for that and skip extra work if not needed?

carve/carve.go
```diff
@@ -0,0 +1,167 @@
...
+	carv2 "github.com/ipld/go-car/v2"
+	"github.com/ipld/go-car/v2/index"
+	"github.com/multiformats/go-multicodec"
+	"github.com/multiformats/go-varint"
+	"github.com/urfave/cli/v2"
+)
+
+func main() {
+	app := &cli.App{
+		Name:  "carve",
```
masih, line 21: As much as I like the name `carve`, my vote is to have a `cli` module and simply call the tool `car` following the naming of similar formats like `tar`

masih (changes requested):
I'd love to reduce the code gymnastics in the CLI and capture some of them in convenient APIs.

Blocker: index marshalling

carve/carve.go
masih: I don't think we can since `BlockReader` does not expose offsets

carve/carve.go
```diff
@@ -0,0 +1,167 @@
...
+				Value:   "CarMultihashIndexSorted",
+			},
+		},
+		Commands: []*cli.Command{
+			{
+				Name:    "index",
+				Aliases: []string{"i"},
+				Usage:   "write out the car with an index",
+				Action: func(c *cli.Context) error {
+					r, err := carv2.OpenReader(c.Args().Get(0))
```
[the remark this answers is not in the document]
willscott: i think this is fine as a positional argument

carve/carve.go
willscott: i think in this context i'm fine without that optimization

masih (approved):

willscott:
I'm noting that running:
```
go run ./cmd/car i ./testdata/sample-v1.car > x.car
```
from the `v2` directory results in the error:
```
expected 1 as the cid version number, got: 4018
```

I wonder if that means this isn't striding over blocks properly / ends up misaligned in the v1 case

PR #220, opened 2021-09-07
stop allocating on every byte read (for CIDs and varints)

mvdan:
(see commit messages - please do not squash)

Using ReadBlocks as an example, we have 30% fewer allocs and 7% less cpu usage.

https://github.com/ipfs/go-cid/pull/132 does another 11% and 3%, respectively.


masih (approved):
thank you 🍻

Issue #95, opened 2021-06-22
Make CAR v2 index iterable without having to read it all into memory


masih:
Ideally, the index would not have to be loaded into memory to be iterated, like [this](https://github.com/filecoin-project/dagstore/pull/2/files#diff-b0929dacd4f1fc192cd308c1a9db2e85154ea742ccec315518e659d7f6a3efc7R347).

See:
* https://github.com/filecoin-project/dagstore/pull/2#discussion_r656358001

masih:
@aarshkshah1992  further to our discussion deferring until requirements are better understood. 

In the meantime looking at the comments on the initial design document, to satisfy iteration over offsets in the index using something like `ForEach(func(c cid.Cid, offset int64) (ok bool, err error)) error` one can:
1. open a CARv2 blockstore
2. get the channel of all keys using `blockstore.AllKeysChan`
3. call `index.Get` to get the offset of the block relative to CARv1 payload.

Key thing to point out here is that `SortedIndex` stores the CID digest only. So providing this style of iteration using index alone could be a bit tricky. However, straightforward to expose on blockstore API _if_ needed. Would be fantastic to understand the actual usecase before delivering this.

gammazero:
@masih 

One reason that the index is loaded into memory is so that it can be binary searched efficiently. May I suggest sorting the index in Eytzinger order so that a binary search can be performed by seeking ahead to the next chunk of the index to read in the search. See the following for an explanation of Eytzinger binary search:
- https://medium.com/swlh/binary-search-vs-eytzinger-order-301f0a9a797d
- https://algorithmica.org/en/eytzinger

PR #221, opened 2021-09-07
Expose the ability to iterate over records in `MultihasIndexSorted`

masih:
Implement API that allows iteration over multihashes and offsets in
`MultihasIndexSorted`. The API is specific to this index type;
therefore, the type is now exported along with its constructor funtion.

Write test that asserts `GetAll` and `ForEach` behave consistently.

Relates to #95

### DIFF
--- a/v2/index/indexsorted.go
+++ b/v2/index/indexsorted.go
@@ -13,6 +13,8 @@ import (
 	"github.com/multiformats/go-multihash"
 )
 
+var _ Index = (*multiWidthIndex)(nil)
+
 type (
 	digestRecord struct {
 		digest []byte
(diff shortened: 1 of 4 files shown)


v2/index/indexsorted.go
```diff
@@ -120,6 +122,21 @@ func (s *singleWidthIndex) Load(items []Record) error {
 	return nil
 }
 
+func (s *singleWidthIndex) forEachDigest(f func(digest []byte, offset uint64) error) error {
+	segmentCount := len(s.index) / int(s.width)
+	for i := 0; i < segmentCount; i++ {
+		digestStart := i * int(s.width)
```
willscott, line 128: should we make sure these are int64 lengths?

willscott (approved):

PR #225, opened 2021-09-08
Make `MultihashIndexSorted` the default index codec for CARv2

### LINKED ISSUE #224: Make `MultihashIndexSorted` the default index type in CARv2
Now that there is the new `MultihashIndexSorted` type, make it the default index codec type across the CARv2 module.

Note, the sample CAR files in `testdata` use `IndexSorted`. They need to be re-named to signal what index type they contain.  Tests need to be updated too.

masih:
Provide an option to choose the index codec when writing a CARv2, with
default value of `multicodec.CarMultihashIndexSorted`.

Implement utility functions to work with options with ability to split
and merge between different flavours to reduce boilerplate code across
the repo.

Re-generate sample files with default index.

Fixes #224


v2/options.go
```diff
@@ -11,6 +13,14 @@ type ReadOptions struct {
 	BlockstoreUseWholeCIDs bool
 }
 
+func NewReadOptions(ropts ...ReadOption) ReadOptions {
```
mvdan, line 16: how about `ApplyReadOptions`, so it's a bit clearer what it does?

v2/options.go
```diff
@@ -26,10 +36,24 @@ var _ ReadWriteOption = ReadOption(nil)
...
 }
 
+// NewWriteOptions instantiates a new WriteOptions and sets the default values for unspecified options.
+func NewWriteOptions(ropts ...WriteOption) WriteOptions {
+	var opts WriteOptions
+	for _, opt := range ropts {
+		opt(&opts)
+	}
+	// Set defaults for zero valued fields.
+	if opts.IndexCodec == multicodec.Code(0) {
```
mvdan, line 51: Doesn't `== 0` work, as an untyped constant?

v2/options.go
```diff
@@ -42,6 +66,33 @@ type ReadWriteOption interface {
...
+func MergeReadWriteOptions(ro []ReadOption, wo []WriteOption) []ReadWriteOption {
+	var rwopts []ReadWriteOption
+	for _, o := range ro {
+		rwopts = append(rwopts, o)
+	}
+	for _, o := range wo {
+		rwopts = append(rwopts, o)
+	}
+	return rwopts
+}
```
mvdan, line 94: hmm, perhaps this is overkill/unnecessary - you never actually merge them. you just convert from one slice type to another. that's three lines of very trivial Go, so I don't think it warrants new API.

mvdan (approved):
SGTM, just minor stuff

v2/index_gen.go
```diff
@@ -5,40 +5,40 @@ import (
...
-	for _, opt := range opts {
-		opt(&o)
+	idx := index.NewMultihashSorted()
+	if err := LoadIndex(idx, v1r, opts...); err != nil {
+		return nil, err
 	}
+	return idx, nil
+}
 
+// LoadIndex populates the given idx with index records generated from the given v1r reader.
```
mvdan, line 27:
```suggestion
// LoadIndex populates idx with index records generated from v1r.
```

v2/index_gen.go
```diff
@@ -48,52 +48,54 @@ func GenerateIndex(v1r io.Reader, opts ...ReadOption) (index.Index, error) {
...
+		return err
 	}
 
-	return idx, nil
+	return nil
 }
 
 // GenerateIndexFromFile walks a car v1 file at the give path and generates an index of cid->byte offset.
 // The index can be stored using index.Save into a file or serialized using index.WriteTo.
+// See GenerateIndex
```
mvdan, line 98: add a period? or finish this sentence

v2/options.go
```diff
@@ -11,6 +13,15 @@ type ReadOptions struct {
 	BlockstoreUseWholeCIDs bool
 }
 
+// ApplyReadOptions applies the given ropts and returns the resulting ReadOptions.
```
mvdan, line 16:
```suggestion
// ApplyReadOptions applies ropts and returns the resulting ReadOptions.
```

v2/options.go
```diff
@@ -42,6 +67,21 @@ type ReadWriteOption interface {
 	readWriteOption()
 }
 
+// SplitReadWriteOptions splits the rwopts by type into ReadOption and WriteOption slices.
+func SplitReadWriteOptions(rwopts ...ReadWriteOption) ([]ReadOption, []WriteOption) {
```
mvdan, line 71: I'm wondering if these should go in an internal package - there's no good reason to export them, I think. Most users should just give vararg options.

mvdan (approved):

v2/options.go
masih: Move to internal causes cycle in package imports. dropped it in favour of duplicate code for now.

PR #226, opened 2021-09-08
Add `car split` command

willscott:
Splits the index off of a carv2 into a separate file.


v2/cmd/car/split.go
```diff
@@ -0,0 +1,25 @@
...
+	r, err := carv2.OpenReader(c.Args().Get(0))
+	if err != nil {
+		return err
+	}
+	defer r.Close()
+
+	if !r.Header.HasIndex() {
+		return fmt.Errorf("no index present")
+	}
+	_, err = io.Copy(os.Stdout, r.IndexReader())
```
mvdan, line 23: consider making this a second argument on disk, because printing to a terminal will almost never be a good idea.

mvdan (approved):

PR #227, opened 2021-09-10
Add `list` and `filter` commands

willscott:
example usage:

```
car list test.car | head -n 50 | car filter test.car trimmed.car
```

### DIFF
--- a/v2/cmd/car/filter.go
+++ b/v2/cmd/car/filter.go
@@ -1,0 +1,100 @@
+package main
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"os"
+	"strings"
+
+	"github.com/ipfs/go-cid"
+	carv2 "github.com/ipld/go-car/v2"
+	"github.com/ipld/go-car/v2/blockstore"
+	icarv1 "github.com/ipld/go-car/v2/internal/carv1"
+	"github.com/urfave/cli/v2"
+)
+
+// FilterCar is a command to select a subset of a car by CID.
+func FilterCar(c *cli.Context) error {
+	r, err := carv2.OpenReader(c.Args().Get(0))
+	if err != nil {
+		return err
+	}
+	defer r.Close()
+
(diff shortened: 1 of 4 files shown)


v2/cmd/car/list.go
```diff
@@ -0,0 +1,47 @@
...
+
+	for {
+		blk, err := rd.Next()
+		if err != nil {
+			if err == io.EOF {
+				break
+			}
+			return err
+		}
+		outStream.WriteString(fmt.Sprintf("%s\n", blk.Cid()))
```
mvdan, line 43:
```suggestion
		fmt.Fprintf(outStream, "%s\n", blk.Cid())
```

v2/cmd/car/list.go
```diff
@@ -0,0 +1,47 @@
...
+	outStream := os.Stdout
+	if c.Args().Len() >= 2 {
+		outStream, err = os.Create(c.Args().Get(1))
+		if err != nil {
+			return err
+		}
+	}
+	defer outStream.Close()
+
+	rd, err := icarv1.NewCarReader(r.DataReader())
```
mvdan, line 30: use https://pkg.go.dev/github.com/ipld/go-car/v2@master#BlockReader? it was written precisely for this :) and it even supports both carv1 and carv2.

v2/cmd/car/filter.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+		defer inStream.Close()
+	}
+	cidList, err := parseCIDS(inStream)
+	if err != nil {
+		return err
+	}
+	fmt.Printf("filtering to %d cids\n", len(cidList))
+
+	cidMap := make(map[cid.Cid]struct{})
```
mvdan, line 52: consider making parseCIDs build a map directly

mvdan: also I wonder if you should at least print a warning on duplicate CIDs

mvdan (approved):

PR #228, opened 2021-09-10
add interface describing iteration

willscott:


masih (approved):

mvdan (approved):
I might have gone with a `func(...) bool` callback, returning false to stop iteration, but I assume this is just the interface over the existing implementation.

I think we should also assert that existing implementations implement it:

    var _ IterableIndex = (*sortedIndex)(nil)

PR #230, opened 2021-09-12
add `get block` to car cli

willscott:


v2/cmd/car/get.go
```diff
@@ -0,0 +1,34 @@
...
+	if c.Args().Len() < 2 {
+		return fmt.Errorf("usage: car get-block <file.car> <block cid>")
+	}
+
+	bs, err := blockstore.OpenReadOnly(c.Args().Get(0))
+	if err != nil {
+		return err
+	}
+
+	// string to car
```
masih, line 22:
```suggestion
	// string to CID
```

v2/cmd/car/get.go
```diff
@@ -0,0 +1,34 @@
...
+	blkCid, err := cid.Parse(c.Args().Get(1))
+	if err != nil {
+		return err
+	}
+
+	blk, err := bs.Get(blkCid)
+	if err != nil {
+		return err
+	}
+	fmt.Printf(string(blk.RawData()))
```
masih, line 32: Could this mess up the terminal in case we get special characters by direct raw data to string conversion?

masih (approved):

PR #231, opened 2021-09-12
Separate CLI to separate module

willscott:
* move to `go get github.com/ipld/go-car/cmd/car`
* remove use of `v2/internal` packages


cmd/car/filter.go
```diff
@@ -49,7 +49,7 @@ func FilterCar(c *cli.Context) error {
 	}
 	fmt.Printf("filtering to %d cids\n", len(cidMap))
 
-	rd, err := icarv1.NewCarReader(r.DataReader())
+	rd, err := carv1.NewCarReader(r.DataReader())
```
masih, line 52: Unrelated to this PR: I think instead of opening a `carv2.Reader` then opening a `carv1.Reader` from the data payload, you can directly use `carv2.BlockReader` which gives you an iterator over CAR blocks may it be a v2 or v2 along with access to roots etc.

masih (approved):

PR #232, opened 2021-09-12
Add `car get-dag` command

willscott:
* cleanup `filter` subcommand
* `get-dag` will take a root CID and selector, and walk the provided dag into a new destination car.


cmd/car/get.go
```diff
@@ -43,3 +54,101 @@ func GetCarBlock(c *cli.Context) error {
...
+		return err
+	}
+
+	ls := cidlink.DefaultLinkSystem()
+	ls.StorageReadOpener = func(_ linking.LinkContext, l datamodel.Link) (io.Reader, error) {
+		if cl, ok := l.(*cidlink.Link); ok {
+			blk, err := bs.Get(cl.Cid)
+			if err != nil {
+				if err == ipfsbs.ErrNotFound {
+					if c.IsSet("strict") && c.Bool("strict") {
```
masih, line 86: Would it suffice to get the value straight up? Explicit or implicit value of `false` would have the same effect I think.

masih (approved):
nice

PR #233, opened 2021-09-13
integrate `car/` cli into `cmd/car`

willscott:
* Flesh out the verify command with v2 checks
* The `list` command is already present
* Removes the `header` command that previously printed out a json representation of a header. It's unclear what we got from that and what we want to have that do with v2 files. I believe the functionality should be largely maintained, and would prefer to either print relevant data as part of `verify` or add an `inspect` command that provides info on both the v2 and v1 metadata if we decide there is important functionality to maintain there.


rvagg (approved):
👌 ftr I'm not convinced of the utility of `verify` either, your call on what to remove, as long as we get rid of the old `main` then this is good by me

willscott:
I think i imagine keeping verify as a baseline for testing that mutation sub-commands generate reasonable output

mvdan (approved):

PR #234, opened 2021-09-13
cmd/car: add first testscript tests

mvdan:
(see commit message)


mvdan:
Also, this brings us up to ~16% coverage of cmd/car itself:

```
$ go test -coverprofile=/tmp/f.html
PASS
coverage: 0.0% of statements
total coverage: 16.1% of statements
ok  	github.com/ipld/go-car/cmd/car	0.033s
```

Note that the first coverage result is smaller, because it doesn't include the merged coverage profiles from the "execs" of the car tool in each script yet.

cmd/car/testdata/script/get-block.txt
```diff
@@ -0,0 +1,19 @@
+env SAMPLE_CID='bafy2bzacebohz654namrgmwjjx4qmtwgxixsd7pn4tlanyrc3g3hwj75hlxrw'
+env MISSING_CID='bafy2bzacebohz654namrgmwjjx4qmtwgxixsd7pn4tlanyrc3g3hwj75xxxxx'
+
+# "get-block" on a CARv1 with an output file.
+car get-block ${INPUTS}/sample-v1.car ${SAMPLE_CID} out.block
```
willscott, line 5: can we re-use the sample cars in the other parts of this same repo?

willscott (approved):

cmd/car/car.go
```diff
@@ -79,7 +81,8 @@ func main() {
 
 	err := app.Run(os.Args)
 	if err != nil {
-		log.Fatal(err)
-		os.Exit(1)
+		log.Println(err)
+		return 1
```
masih, line 85: Would it make sense to utilise urfave/cli machinery like [`OsExiter`](https://pkg.go.dev/github.com/urfave/cli/v2#OsExiter) for handling exit codes here? Tests can then override that.

masih (approved):
nice

cmd/car/testdata/script/get-block.txt
mvdan: Unfortunately we cannot, because each module is zipped separately. If we make them share testdata files across module boundaries, then the tests only pass if you've got a full `git clone`.

mvdan: I think this is fine in practice because the library needs lots of different input files to exercise error cases, but I think the CLI should only need 2-3 good input files to test happy paths.

cmd/car/car.go
mvdan: I guess we technically could, but I'm not sure that we should. `func() int` feels much easier than abusing a library's global callback variable :)

mvdan:
Rebased; should be ready to merge.

willscott (approved):

PR #235, opened 2021-09-13
Reproduce "illegal wire type" failure

aarshkshah1992:


mvdan:
I seem to recall we fixed this by adding opt-in support for identity CIDs. Please reopen this or open an issue if I got that wrong. Thanks!

PR #236, opened 2021-09-15
Add test script for car verify

willscott:


mvdan (approved):
SGTM assuming verify prints nothing when it's happy

willscott:
Not to expand scope too much, but here's also a commit to test the filter command.

willscott:
I feel a bit dirty about the need for windows specific line endings. I guess the alternative is that i'd need to have the test figure file embedded in filter.txt split into a separate file, maybe?

mvdan:
It's best to just not let git convert plaintext files to CRLF: https://github.com/mvdan/sh/blob/master/.gitattributes

willscott:
> It's best to just not let git convert plaintext files to CRLF: https://github.com/mvdan/sh/blob/master/.gitattributes

That appears to have worked :+1:

PR #237, opened 2021-09-15
Combine API options for simplicity and logical coherence

masih:
Introduce a simple `Option` type that includes both read and write
options, as well as read-write options. The rationale for this change is
to avoid re-declaring options that apply to both read and write, and
reduce code verbosity when converting from one type to another as part
of API call chains.

Deprecate `ReadOption`, `WriteOption` and `ReadWriteOption` in favour of
the simpler `Option`, and update APIs to use the new type.

Remove now redundant internal utility functions for working with option
types.

Reformat code using `gofumt` utility for consistency.


masih:
@mvdan  @willscott New options coming up re handling of IDENTITY can apply to both read and write. The changes in this PR paves the way for a simpler option handing for the upcoming PR as well as reducing the conversion between types and simpler API interaction.

v2/options.go
```diff
@@ -2,51 +2,57 @@ package car
...
+type ReadOption Option
 
 func (ReadOption) readWriteOption() {}
 
 var _ ReadWriteOption = ReadOption(nil)
 
-// WriteOptions holds the configured options after applying a number of
-// WriteOption funcs.
+// WriteOption hints that an API wants options related only to reading CAR files.
+// Deprecated: use Option instead.
```
mvdan, line 21: I'm not sure I would deprecate ReadOption and WriteOption. If an API *only* reads or *only* writes, taking one of those two will then help make the API clearer, even if it's just for the sake of names and docs.

v2/options.go
```diff
@@ -2,51 +2,57 @@ package car
...
-	}
-	return opts
-}
+var _ ReadWriteOption = Option(nil)
 
-// ReadOption describes an option which affects behavior when parsing CAR files.
-type ReadOption func(*ReadOptions)
+// ReadOption hints that an API wants options related only to reading CAR files.
+// Deprecated: use Option instead.
+type ReadOption Option
```
mvdan, line 14: You want ReadOption, WriteOption, and ReadWriteOption to be aliases now; like

```
type ReadOption = Option
type WriteOption = Option
type ReadWriteOption = Option
```

The reason is to minimize breakage downstream, and to allow some APIs to keep using ReadOption and WriteOption to signal that they don't use all options.

mvdan (approved):
Nice cleanup. Hindsight is 20/20; I think we were right to try to split the options by their read vs write semantics, but it's gotten to a point where the line is too fuzzy and it's causing more harm than good.

masih:
Thanks for the speedy review @mvdan 🚀

mvdan (approved):
:ok_hand:

PR #239, opened 2021-09-17
Implement options to handle `IDENTITY` CIDs gracefully

masih:
Implement two additional options that allow a CARv2 file to 1) include
IDENTNTIY CIDs, and 2) specify a maximum allowed CID length with default
of 2KiB as a sufficiently large default.

Configure ReadWrite blockstore to persist given blocks with IDENTITY
CIDs.

Introduce a new Characteristics filed that signalls whether an index in
a CAR file contains a full catalog of CIDs for backward compatibility
purposes. Note, this is a new addition and will need to be added to the
spec in a separate PR.

Relates to #215

### DIFF
--- a/v2/blockstore/readwrite.go
+++ b/v2/blockstore/readwrite.go
@@ -295,11 +295,23 @@ func (b *ReadWrite) PutMany(blks []blocks.Block) error {
 	for _, bl := range blks {
 		c := bl.Cid()
 
-		// Check for IDENTITY CID. If IDENTITY, ignore and move to the next block.
-		if _, ok, err := isIdentity(c); err != nil {
-			return err
-		} else if ok {
-			continue
+		// If StoreIdentityCIDs option is disabled then treat IDENTITY CIDs like IdStore.
+		if !b.opts.StoreIdentityCIDs {
+			// Check for IDENTITY CID. If IDENTITY, ignore and move to the next block.
+			if _, ok, err := isIdentity(c); err != nil {
+				return err
+			} else if ok {
+				continue
+			}
+		}
+
+		// Check if its size is too big.
+		// If larger than maximum allowed size, return error.
+		// Note, we need to check this regardless of whether we have IDENTITY CID or not.
+		// Since multhihash codes other than IDENTITY can result in large digests.
+		cSize := uint64(len(c.Bytes()))
+		if cSize > b.opts.MaxIndexCidSize {
+			return &carv2.ErrCidTooLarge{MaxSize: b.opts.MaxIndexCidSize, CurrentSize: cSize}
 		}
 
 		if !b.opts.BlockstoreAllowDuplicatePuts {
@@ -351,6 +363,7 @@ func (b *ReadWrite) Finalize() error {
 
 	// TODO check if add index option is set and don't write the index then set index offset to zero.
 	b.header = b.header.WithDataSize(uint64(b.dataWriter.Position()))
+	b.header.Characteristics.SetFullyIndexed(b.opts.StoreIdentityCIDs)
 
 	// Note that we can't use b.Close here, as that tries to grab the same
 	// mutex we're holding here.
--- a/v2/blockstore/readwrite_test.go
+++ b/v2/blockstore/readwrite_test.go
@@ -2,6 +2,7 @@ package blockstore_test
 
 import (
 	"context"
+	"crypto/sha512"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -12,21 +13,19 @@ import (
 	"testing"
 	"time"
 
-	"github.com/ipfs/go-merkledag"
-
-	carv2 "github.com/ipld/go-car/v2"
-	"github.com/ipld/go-car/v2/index"
-	"github.com/stretchr/testify/assert"
-
-	"github.com/multiformats/go-multihash"
-	"github.com/stretchr/testify/require"
-
-	ipfsblockstore "github.com/ipfs/go-ipfs-blockstore"
-	"github.com/ipld/go-car/v2/blockstore"
-
 	blocks "github.com/ipfs/go-block-format"
 	"github.com/ipfs/go-cid"
+	ipfsblockstore "github.com/ipfs/go-ipfs-blockstore"
+	cbor "github.com/ipfs/go-ipld-cbor"
+	"github.com/ipfs/go-merkledag"
+	carv2 "github.com/ipld/go-car/v2"
+	"github.com/ipld/go-car/v2/blockstore"
+	"github.com/ipld/go-car/v2/index"
 	"github.com/ipld/go-car/v2/internal/carv1"
+	"github.com/multiformats/go-multicodec"
+	"github.com/multiformats/go-multihash"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 var (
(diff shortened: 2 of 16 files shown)


v2/car.go
```diff
@@ -43,6 +43,9 @@ type (
 	}
 )
 
+// fullyIndexedCharPos is the position of Characteristics.Hi bit that specifies whether the index is a catalog af all CIDs or not.
+const fullyIndexedCharPos = 0
```
mvdan, line 47: this will largely depend on the spec, but if you're using the lowest bit of the high uint64, then you're assigning one of the two bits right in the middle :)

mvdan: another way to verify this is to e.g. hex encode the characteristic bits, and see that they visually correspond to what you'd expect - e.g. the "fully indexed" bit being the first, thus all the way at the left (highest)

v2/errors.go
```diff
@@ -0,0 +1,28 @@
+package car
+
+import (
+	"fmt"
+)
+
+type (
+	// ErrCidTooLarge signals that a CID is too large to include in CARv2 index.
+	ErrCidTooLarge error
```
mvdan, line 9: not sure I get the point of this error type. Either export errCidTooLarge with its fields, or since NewErrCidTooLarge is exported, or unexport NewErrCidTooLarge and just use `error` types as return parameters.

v2/index_gen.go
```diff
@@ -76,7 +80,13 @@ func LoadIndex(idx index.Index, v1r io.Reader, opts ...Option) error {
 		if err != nil {
 			return err
 		}
-		records = append(records, index.Record{Cid: c, Offset: uint64(sectionOffset)})
+
+		if c.Prefix().MhType != multihash.IDENTITY || o.IncludeIdentityCIDs {
```
mvdan, line 84: nit: put the IncludeIdentityCIDs check first, as it's cheaper/simpler

v2/index/index.go
```diff
@@ -47,6 +45,10 @@ type (
 		Unmarshal(r io.Reader) error
 
 		// Load inserts a number of records into the index.
+		// Note that Index will load all given records. Any filtering of the records such as
+		// exclusion of CIDs with multihash.IDENTITY code must occur prior to calling this function.
+		// Further, the actual information extracted and indexed from the given records entirely
+		// depends on the concrete index implementation.
```
mvdan, line 51: I think that last sentence deserves to be on a separate paragrah and with an example - "for example, some index implementations may only store partial multihashes".

v2/blockstore/readwrite_test.go
```diff
@@ -688,3 +687,165 @@ func TestReadWriteErrorAfterClose(t *testing.T) {
...
+	require.Equal(t, idBlock, gotBlock)
+
+	next, err := reader.Next()
+	require.Equal(t, io.EOF, err)
+	require.Nil(t, next)
+
+	// Assert the id is indexed.
+	r, err := carv2.OpenReader(path)
+	require.NoError(t, err)
+	t.Cleanup(func() { require.NoError(t, r.Close()) })
```
mvdan, line 745: no need to check read-only close errors :wink:

mvdan (approved):

v2/car.go
mvdan: I *might* be wrong on all this, because thinking about bits and endianess is sometimes tricky. All the more reason to visualize how the bits are laid out when encoded via a test.

rvagg: > tricky

An understatement, trying to pin this all down for the HAMT spec was crazy and we've still ended up with differing implementations and the cross-language challenges are a headache.

Arguably using `Hi` is the correct choice here if we're thinking of this as an "ordered" bit field since we're writing it out first. I think even this didn't need to even be a pair of integers but just a `[16]byte` that we manipulate directly (just not with `big.Int` _please!).

I wrote https://github.com/rvagg/bit-sequence to do JS & Go versions of what it means to slice a bit field into sections of integers by arbitrary bit boundaries. It's not quite the same thing here since we're dealing with a single bit, but maybe we could verify that what's being set here is a 1-bit integer at the position in question, and then we'd have a consistent reference to point to at least.

Not something to get too hung up on, however. There's a first-mover advantage here in that what you define now is going to be locked in for any other implementation and they'll (including JS) will just have to roll with that.

v2/blockstore/readwrite_test.go
```diff
@@ -688,3 +687,165 @@ func TestReadWriteErrorAfterClose(t *testing.T) {
...
+	require.True(t, bigCidLen > maxAllowedCidSize)
+
+	bigBlock, err := blocks.NewBlockWithCid(data, bigCid)
+	require.NoError(t, err)
+	err = subject.Put(bigBlock)
+	require.Equal(t, carv2.NewErrCidTooLarge(maxAllowedCidSize, bigCidLen), err)
+}
+
+func TestReadWrite_ReWritingCARv1WithIdentityCidIsIdenticalToOriginalWithOptionsEnabled(t *testing.T) {
+	originalCARv1Path := "../testdata/sample-v1.car"
```
rvagg, line 799: so `sample-v1.car` contains identity CIDs? is there already coverage of the case where this fixture gets written _without_ identity CIDs?

v2/index_gen.go
```diff
@@ -76,7 +80,13 @@ func LoadIndex(idx index.Index, v1r io.Reader, opts ...Option) error {
...
+			if uint64(cidLen) > o.MaxIndexCidSize {
```
rvagg, line 85: coverage of this case might be nice - it could even fit in one of the existing tests here without needing a special test just for it

rvagg:
The maximum of 2K is somewhat arbitrary isn't it? Not disagreeing with it, it feels arbitrary no matter how you slice it, but if we have a rationale then that might be good to include in a note.

By locking it in a default here as 2K we're probably going to be pushing that fairly broadly through the stack if the CARv2 writer interface becomes the default for creating CARs. Which is probably good we have a proper limiting condition, just with possible ramifications.

I might have to consider getting something similar in to the JavaScript CAR because we now have a lot of user data being generated that way, it's possible we get large identity CIDs slipping in there and only being caught if/when they hit this code in Filecoin.

v2/car.go
mvdan: +1 to making characteristics a byte array, actually. If you remove the complexity of having to think about endianness, I think our code and docs would get a lot simpler.

The spec allows it - it only says it's 128 bits or 16 bytes, it doesn't require any representation in memory.

Technically a breaking change in terms of Go APIs, but I think we're in the clear given how the characteristics bits are entirely unused right now.

v2/index_gen.go
masih: There is a test for large CIDs already in this PR [here](https://github.com/ipld/go-car/blob/9f9f687b3606dae462007d93cc0e8684129435f5/v2/blockstore/readwrite_test.go#L778-L795). Do you think it's missing something?

v2/blockstore/readwrite_test.go
masih: > so `sample-v1.car` contains identity CIDs? 

Yes it does.

> is there already coverage of the case where this fixture gets written _without_ identity CIDs?

Not sure; great shout. Adding. Thanks 🍻

v2/car.go
masih: > maybe we could verify that what's being set here is a 1-bit integer at the position in question, and then we'd have a consistent reference to point to at least.

Does [this test](https://github.com/ipld/go-car/blob/9f9f687b3606dae462007d93cc0e8684129435f5/v2/car_test.go#L206-L221) do the trick on verification?


> +1 to making characteristics a byte array, actually.

I think it was a byte array at first but we settled on `Hi` `Low` int64; I am failing to find reference discussions apologies. 
Happy to change it in a separate PR if one of you gents would kindly volunteer to write up an issue?

masih:
> The maximum of 2K is somewhat arbitrary isn't it? Not disagreeing with it, it feels arbitrary no matter how you slice it, but if we have a rationale then that might be good to include in a note.
> 
> By locking it in a default here as 2K we're probably going to be pushing that fairly broadly through the stack if the CARv2 writer interface becomes the default for creating CARs. Which is probably good we have a proper limiting condition, just with possible ramifications.
> 
> I might have to consider getting something similar in to the JavaScript CAR because we now have a lot of user data being generated that way, it's possible we get large identity CIDs slipping in there and only being caught if/when they hit this code in Filecoin.

It is mostly arbitrary indeed; the rationale was to pick something that is somewhat future proof and unlikely to be hit in reasonable use cases today. Any suggestions for a better value is more than welcomed. I know @aschmahmann pointed out a limit of `80` in [KAD DHT](https://github.com/libp2p/go-libp2p-kad-dht/blob/8d8f18fe409c2aed58e34d8f1ca067ba1c27f99c/handlers.go#L312)? The key thing to point out here though, is that, this is an **option default value** and not a hard limit; i.e. option can be overridden and set to all sorts of unreasonable values.

Thoughts?

v2/blockstore/readwrite_test.go
masih: Looks like there is yes. The test [here](https://github.com/ipld/go-car/blob/9f9f687b3606dae462007d93cc0e8684129435f5/v2/blockstore/readwrite_test.go#L50-L126) asserts that when the option is not set the resulting CARv2, written using `ReadWrite` blockstore does not include `IDENTITY` CIDs despite `Put`ing them.

Please shout if I have missed anything.

v2/car.go
masih: Thanks for capturing this @mvdan https://github.com/ipld/go-car/issues/240

PR updated to use the left most bit in `Hi`

mvdan (approved):
Nice work!

v2/options.go
```diff
@@ -25,6 +28,8 @@ type Options struct {
 	IndexPadding           uint64
 	IndexCodec             multicodec.Code
 	ZeroLengthSectionAsEOF bool
+	MaxIndexCidSize        uint64
```
ribasushi, line 31: `Size` vs `Length` in the same optionset strike me as a bit odd. No hard preference, just raising it...

v2/index/index.go
```diff
@@ -61,7 +65,7 @@ type (
 	}
 
 	// IterableIndex extends Index in cases where the Index is able to
-	// provide an iterator for getting the list of all entries in the
+	// provide an iterator for getting the list of all multihashes in the
 	// index.
 	IterableIndex interface {
 		Index
```
ribasushi, line 71: Here and a few lines down - probably need to clarify what happens when there are multiple offsets for the same MH...

v2/blockstore/readwrite_test.go
```diff
@@ -688,3 +687,165 @@ func TestReadWriteErrorAfterClose(t *testing.T) {
...
+	require.NoError(t, err)
+	t.Cleanup(func() { require.NoError(t, v2r.Close()) })
+
+	// Assert characteristics bit is set.
+	require.True(t, v2r.Header.Characteristics.IsFullyIndexed())
+
+	// Assert original CARv1 and generated innter CARv1 payload have the same SHA512 hash
+	// Note, we hash instead of comparing bytes to avoid excessive memory usage when sample CARv1 is large.
+
+	hasher := sha512.New()
```
ribasushi, line 837: Interesting choice ;)

v2/blockstore/readwrite.go
```diff
@@ -295,11 +295,23 @@ func (b *ReadWrite) PutMany(blks []blocks.Block) error {
...
 		c := bl.Cid()
 
-		// Check for IDENTITY CID. If IDENTITY, ignore and move to the next block.
-		if _, ok, err := isIdentity(c); err != nil {
-			return err
-		} else if ok {
-			continue
+		// If IncludeIdentityCIDs option is disabled then treat IDENTITY CIDs like IdStore.
+		if !b.opts.IncludeIdentityCIDs {
+			// Check for IDENTITY CID. If IDENTITY, ignore and move to the next block.
```
ribasushi, line 300: I wonder if you really want to convey `IncludeIden...` vs what actually happens: `StoreIden...`

ribasushi (approved):
Solid wok overall, left a number of comments regarding naming/clarifications

None of the remarks are showstoppers, take them or leave them: it's good either way.

v2/options.go
masih: Thanks for pointing this out. `ZeroLength*` is the odd one out; the rest of the codebase uses `Size` consistently. Will leave this as is and consider renaming on a separate PR

v2/index_gen.go
rvagg: yeah, it's not using the default (common) path but setting an unrealistic max to break, that was my thought
but this is a _nice to have_, not a demand for busy-work if you're comfortable with this as is

masih: Thanks for pointing this out; captured https://github.com/ipld/go-car/issues/241

PR #248, opened 2021-09-30
blockstore: OpenReadWrite should not modify if it refuses to resume

mvdan:
(see commit message)

### DIFF
--- a/v2/blockstore/readwrite_test.go
+++ b/v2/blockstore/readwrite_test.go
@@ -600,17 +600,39 @@ func TestReadWriteResumptionFromNonV2FileIsError(t *testing.T) {
 	require.Nil(t, subject)
 }
 
+func TestReadWriteResumptionMismatchingRootsIsError(t *testing.T) {
+	tmpPath := requireTmpCopy(t, "../testdata/sample-wrapped-v2.car")
+
+	origContent, err := ioutil.ReadFile(tmpPath)
+	require.NoError(t, err)
+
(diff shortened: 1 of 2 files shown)


mvdan:
I ended up splitting the "temproary file" idea mentioned in the commit message into https://github.com/ipld/go-car/issues/249.

v2/blockstore/readwrite_test.go
```diff
@@ -600,17 +600,39 @@ func TestReadWriteResumptionFromNonV2FileIsError(t *testing.T) {
...
+	// If the sizes mismatch, printing a huge diff would not help us.
+	require.Equal(t, len(origContent), len(newContent))
+	require.Equal(t, origContent, newContent)
+}
+
 func requireTmpCopy(t *testing.T, src string) string {
 	srcF, err := os.Open(src)
 	require.NoError(t, err)
-	t.Cleanup(func() { require.NoError(t, srcF.Close()) })
+	defer func() { require.NoError(t, srcF.Close()) }()
```
masih, line 628: I am curious; why? I would say `t.Cleanup` is clearer unless it's functionally the incorrect thing to use?

masih (approved):

v2/blockstore/readwrite_test.go
mvdan: Oh, yes, I think Cleanup is the wrong one. If you want to do a temp copy of a file on disk, you want to close both source and dest as soon as the copy is done (when the func returns), not when the entire test finishes.

Issue #196, opened 2021-08-02
carv2/ReadWriteBlockstore: support deferred root CIDs


raulk:
`blockstore.OpenReadWrite` requires providing the root CIDs when creating a new blockstore. This design inhibits the ability to use a CARv2 blockstore as the target of a streaming merkle DAG formation like UnixFS, as the root CID is not known beforehand.

We could work around this situation by supplying a placeholder root CID, and once the blockstore is finalized, we could go back and replace those bytes in the header.

Unfortunately, the library doesn't provide APIs to do that safely and without making assumptions about the underlying format, or breaking abstractions.

Some ideas:

1. Allow `blockstore.OpenReadWrite` to take a `[]struct{cid.Builder, func() (cid.Cid, error)}`.
    - Use the `cid.Builder`s to compute the length of each CID, and preallocate those bytes in the header.
    - On Finalize, call each function to get the actual CID to replace it in the header.
2. Simpler: allow the user to specify a number of bytes to preallocate, and provide primitives to update a CAR header.

masih:
Implementation in lotus [here](https://github.com/filecoin-project/lotus/pull/7079/files#diff-b947764ffb1599ab13eb67261ce286ab799a34ed90124c1d1352b1a48fb1766aR642-R691)

rvagg:
@masih can we close this out? is what we have enough to cover this for now or should we label this as an actual TODO for API improvement?

masih:
I think this issue is worth resolving. The API we have just now doesn't exactly cover the ask here and I think it is a useful thing to have.

MichaelMure:
In a completely different direction, have you considered writing the header+index at the end of the file, [like the zip format does](https://users.cs.jmu.edu/buchhofp/forensics/formats/pkzip.html)? This has several benefit:
- no assumption required about the header/index size
- way easier to build the index and root CID list while writing the blocks, just append at the end
- way easier to append new blocks on an already finalized file: read the previous header/index, overwrite those with the new blocks, write the new header/index again

rvagg:
[...]

PR #250, opened 2021-09-30
Implement API to allow replacing root CIDs in a CARv1 or CARv2

### LINKED ISSUE #245: Better ergonomics around updating car root.
The `v2/blockstore` package requires the caller to specify the car root `[]cid.Cid` on opening the file.
In the common case, the caller will not know their intended root until the end of writing.

This currently seems to require manual file editing by the caller. I've tried finalizing, and the re-opening via
```golang
cdest, err = blockstore.OpenReadWrite(c.String("file"), []cid.Cid{root})
```
specifying the actual root CID, where on the initial write I instead generate a proxy CID that I know has the same length as the final CID will. This fails, telling me `cannot resume on file with mismatching data header`. Forcing me to manually go in and edit the root unsupported seems like a failing in this library interface.

masih:
Implement an API that allows a caller to replace root CIDs in an
existing CAR file, may it be v1 or v2, as long as the resulting
serialized header is of identical size to the existing header.

Assert that the new API works in a variety of CARv1 and CARv2 files
along with failure scenarios.

Fixes #245

### DIFF
--- a/v2/writer.go
+++ b/v2/writer.go
@@ -1,12 +1,15 @@
 package car
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"io"
 	"os"
 
+	"github.com/ipfs/go-cid"
 	"github.com/ipld/go-car/v2/index"
+	"github.com/ipld/go-car/v2/internal/carv1"
 	internalio "github.com/ipld/go-car/v2/internal/io"
 )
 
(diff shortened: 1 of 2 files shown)


mvdan:
I think having it take a path on disk is fair - we only exposed `OpenReadWrite` after all, and no variant that attempts to use io interfaces, because we just need too many methods.

And if we want to later on expose a `FinalizeWithRoots`, we could also do that - and it would internally share the same implementation as `ReplaceRootsInFile`, perhaps via an unexported API that works on `*os.File`.

masih:
> And if we want to later on expose a `FinalizeWithRoots`, we could also do that

That was my thought too. I think we probably need both: Basic APIs to work with CAR files, and equivalent functionality used through blockstore.

I think we have a separate ticket already for blockstore support: #196

v2/writer_test.go
```diff
@@ -131,3 +131,150 @@ func assertAddNodes(t *testing.T, adder format.NodeAdder, nds ...format.Node) {
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// Make a copy of input files to preserve original for comparison.
+			// This also avoids modification files in testdata.
+			tmpCopy := requireTmpCopy(t, tt.path)
+			err := ReplaceRootsInFile(tmpCopy, tt.roots)
+			if tt.wantErr {
+				require.Error(t, err, "ReplaceRootsInFile() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			} else {
```
mvdan, line 216: drop this else and unindent? since the other branch returns anyway

v2/writer_test.go
```diff
@@ -131,3 +131,150 @@ func assertAddNodes(t *testing.T, adder format.NodeAdder, nds ...format.Node) {
...
+				require.NoError(t, err)
+
+				original, err := os.Open(tt.path)
+				require.NoError(t, err)
+				t.Cleanup(func() { require.NoError(t, original.Close()) })
+
+				target, err := os.Open(tmpCopy)
+				require.NoError(t, err)
+				t.Cleanup(func() { require.NoError(t, target.Close()) })
```
mvdan, line 225: might as well just defer :) IMO Cleanup is only useful in test helper funcs, where defer can't be used.

v2/writer_test.go
```diff
@@ -131,3 +131,150 @@ func assertAddNodes(t *testing.T, adder format.NodeAdder, nds ...format.Node) {
...
+func requireDecodedCid(t *testing.T, s string) cid.Cid {
+	decoded, err := cid.Decode(s)
+	require.NoError(t, err)
+	return decoded
+}
+
+func requireTmpCopy(t *testing.T, src string) string {
+	srcF, err := os.Open(src)
+	require.NoError(t, err)
+	t.Cleanup(func() { require.NoError(t, srcF.Close()) })
```
mvdan, line 268: like in my other PR, I think these need to be defers

v2/writer.go
```diff
@@ -210,3 +214,97 @@ func AttachIndex(path string, idx index.Index, offset uint64) error {
...
+
+	var currentSize int64
+	var newHeaderOffset int64
+	switch header.Version {
+	case 1:
+		// When the given file is a CARv1 :
+		// 1. The offset at which the new header should be written is zero (newHeaderOffset = 0)
+		// 2. The current header size is equal to the number of bytes read, and
+		//
+		// Note that we explicitly avoid using carv1.HeaderSize to deremine the current header size.
```
mvdan, line 250: typo: determine

v2/writer.go
```diff
@@ -210,3 +214,97 @@ func AttachIndex(path string, idx index.Index, offset uint64) error {
...
+		newHeaderOffset = int64(v2h.DataOffset)
+		if _, err = f.Seek(newHeaderOffset, io.SeekStart); err != nil {
+			return err
+		}
+		var innerSize uint64
+		innerSize, err = varint.ReadUvarint(internalio.ToByteReader(f))
+		if err != nil {
+			return err
+		}
+		currentSize = int64(innerSize + uint64(varint.UvarintSize(innerSize)))
```
mvdan, line 279: I wonder if it would be simpler to just do the same logic as in the carv1? after all, we just need to do exactly the same stuff, but with the inner section if it's a carv2.

mvdan: another reason to always read the "true" inner carv1 header is to ensure that it's valid. right now, if a carv2 has an inner carv1 header with version=5, you'll happily replace it with version=1 without erroring. instead we should likely complain.

v2/writer_test.go
```diff
@@ -131,3 +131,150 @@ func assertAddNodes(t *testing.T, adder format.NodeAdder, nds ...format.Node) {
 		assert.NoError(t, adder.Add(context.Background(), nd))
 	}
 }
+
+func TestReplaceRootsInFile(t *testing.T) {
+	tests := []struct {
+		name    string
+		path    string
+		roots   []cid.Cid
+		wantErr bool
```
mvdan, line 140: nit: I'd make this a string and do a `strings.Contains`, to have a better idea of how each test case is expected to fail.

I know that comparing error strings is generally a bad thing, but if you're testing your own code, I think doing that is pretty reasonable - and not prone to breakages as long as you check for substrings rather than entire strings.

willscott (approved):
we could potentially get rid of the comment saying it doesn't shift padding - i'd prefer that to show up in this method once supported, rather than maintaining two different exposed APIs for it.

mvdan (approved):

mvdan:
> i'd prefer that to show up in this method once supported, rather than maintaining two different exposed APIs for it.

My only worry is that it would end up being an API that might be very cheap (just rewrite the first kilobyte of data) versus very expensive (copy potentially gigabytes of data, depending on how large the CAR is).

masih:
I think what will is taking about is to use the padding after CARv2 header, before inner CARv1 to grow roots when needed. That should still be cheap for CARv2 files and it just means `DataOffset` in CARv2 header needs to be updated.

mvdan:
Ah, that's fair. As long as it's equally cheap :)

PR #251, opened 2021-10-01
Expose selector traversal options for SelectiveCar

rvagg:
Expose `Options` for `NewSelectiveCar()` for two new features available in go-ipld-prime:

1. `traversal.LinkVisitOnlyOnce` which I've called `TraverseLinksOnlyOnce` here - defaulting to `false`
2. `traversal.Budget#LinkBudget` which I've called `MaxTraversalLinks` here - defaulting to _disabled_

I've gone with the same `Options` pattern as in v2 and taken @masih's advice here to keep it generic (rather than use `SelectiveCarOptions`) so the same pattern can be used for the non-selective CAR usage functions. We'll just have to be clear about what options apply to what functions I suppose.

I've put `Traverse` in both option names for this reason - to try and make it clear that they are for selector traversal. But, in the future we could retire the merkledag `Walk` use here and replace it with ipld-prime's traversal (but with `LinkVisitOnlyOnce` enabled by default which would make it equivalent in functionality). So both of these options would make sense for that case too. I'm open to critique of the naming if anyone has stronger opinions.

I also extracted `TestRoundtripSelective` to the new `selectivecar_test.go`, but the other two top-level tests in there are new.

### DIFF
--- a/options.go
+++ b/options.go
@@ -2,7 +2,7 @@ package car
 
 import "math"
 
-// Options holds the configured options after applying a number of
+// options holds the configured options after applying a number of
 // Option funcs.
 type options struct {
 	TraverseLinksOnlyOnce bool
@@ -38,7 +38,7 @@ func MaxTraversalLinks(MaxTraversalLinks uint64) Option {
 	}
 }
 
-// ApplyOptions applies given opts and returns the resulting Options.
+// applyOptions applies given opts and returns the resulting options.
 func applyOptions(opt ...Option) options {
 	opts := options{
 		TraverseLinksOnlyOnce: false,         // default: recurse until exhausted
--- a/options.go
+++ b/options.go
@@ -1,0 +1,56 @@
+package car
+
+import "math"
+
+// Options holds the configured options after applying a number of
+// Option funcs.
+//
+// This type should not be used directly by end users; it's only exposed as a
+// side effect of Option.
+type Options struct {
+	TraverseLinksOnlyOnce bool
+	MaxTraversalLinks     uint64
+}
+
+// Option describes an option which affects behavior when
+// interacting with the  interface.
+type Option func(*Options)
+
+// TraverseLinksOnlyOnce prevents the traversal engine from repeatedly visiting
+// the same links more than once.
+//
+// This can be an efficient strategy for an exhaustive selector where it's known
+// that repeat visits won't impact the completeness of execution. However it
+// should be used with caution with most other selectors as repeat visits of
+// links for different reasons during selector execution can be valid and
+// necessary to perform full traversal.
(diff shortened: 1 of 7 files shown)


options.go
```diff
@@ -0,0 +1,56 @@
...
+		sco.MaxTraversalLinks = MaxTraversalLinks
+	}
+}
+
+// ApplyOptions applies given opts and returns the resulting Options.
+// This function should not be used directly by end users; it's only exposed as a
+// side effect of Option.
+func ApplyOptions(opt ...Option) Options {
+	opts := Options{
+		TraverseLinksOnlyOnce: false,         // default: recurse until exhausted
```
masih, line 49: I'd certainly defer to your judgment on the default value for this option. 
Thinking out loud I would have thought a typical user would want to set this to `true`, and maybe default `true` (or changing the name so that zero value translates to traversing only once) would be more reasonable?

options.go
```diff
@@ -0,0 +1,56 @@
...
+// execution when building a SelectiveCar.
+func MaxTraversalLinks(MaxTraversalLinks uint64) Option {
+	return func(sco *Options) {
+		sco.MaxTraversalLinks = MaxTraversalLinks
+	}
+}
+
+// ApplyOptions applies given opts and returns the resulting Options.
+// This function should not be used directly by end users; it's only exposed as a
+// side effect of Option.
```
masih, line 46: non-blocker: Considering the go-car/v0 package structure I _think_ we don't have to expose this yet?

selectivecar_test.go
```diff
@@ -0,0 +1,227 @@
...
+
+type countingReadStore struct {
+	bs    car.ReadStore
+	count int
+}
+
+func (rs *countingReadStore) Get(c cid.Cid) (blocks.Block, error) {
+	rs.count++
+	return rs.bs.Get(c)
+}
```
masih, line 227: Thank you for moving these tests 🍻

masih (approved):

options.go
rvagg: 👍 can you look at 3ff3c49 please and let me know if I'm getting the visibility right for the different pieces?

options.go
masih: I see great points thank you

options.go
```diff
@@ -0,0 +1,51 @@
+package car
+
+import "math"
+
+// Options holds the configured options after applying a number of
```
masih, line 5:
```suggestion
// options holds the configured options after applying a number of
```

masih (approved):

options.go
masih: LGTM thanks for reflecting on my comment

PR #255, opened 2021-10-07
add ZeroLengthSectionAsEOF option to the v0 module

mvdan:
(see commit message)

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -2,60 +2,35 @@ package util
 
 import (
 	"bufio"
-	"bytes"
 	"encoding/binary"
 	"fmt"
 	"io"
 
 	cid "github.com/ipfs/go-cid"
-	mh "github.com/multiformats/go-multihash"
 )
-
-var cidv0Pref = []byte{0x12, 0x20}
 
 type BytesReader interface {
 	io.Reader
 	io.ByteReader
 }
 
-// TODO: this belongs in the go-cid package
+// Deprecated: use go-cid.CidFromBytes.
 func ReadCid(buf []byte) (cid.Cid, int, error) {
-	if bytes.Equal(buf[:2], cidv0Pref) {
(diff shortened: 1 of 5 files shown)


mvdan:
FYI @aarshkshah1992 @dirkmc

util/util.go
```diff
@@ -2,61 +2,36 @@ package util
...
 func ReadNode(br *bufio.Reader) (cid.Cid, []byte, error) {
 	data, err := LdRead(br)
 	if err != nil {
 		return cid.Cid{}, nil, err
 	}
+	// ReadCid used to panic or error on null padding.
+	// Instead, return a sentinel error to let the user decide what to do.
+	if len(data) == 0 {
+		return cid.Cid{}, nil, ErrZeroLengthSection
+	}
```
masih, line 34: I am curious why not take option here similar [to internal fork in v2](https://github.com/ipld/go-car/blob/master/v2/internal/carv1/util/util.go#L18)?

util/util.go
```diff
@@ -2,61 +2,36 @@ package util
...
+
 func ReadNode(br *bufio.Reader) (cid.Cid, []byte, error) {
 	data, err := LdRead(br)
 	if err != nil {
 		return cid.Cid{}, nil, err
 	}
+	// ReadCid used to panic or error on null padding.
+	// Instead, return a sentinel error to let the user decide what to do.
+	if len(data) == 0 {
+		return cid.Cid{}, nil, ErrZeroLengthSection
```
willscott, line 33: cid.Undef ?

willscott (approved):

util/util.go
mvdan: I'm just following the rest of the code :) I don't mind one way or another, but it seems that consistency is best, and the cleanup seems a bit out of place for a bugfix

PR #256, opened 2021-10-11
iterative write

coryschwartz:
This edit was made for https://github.com/ipfs/go-ipfs/pull/8503


aschmahmann:
It seems like this is not needed anymore. If that's incorrect feel free to reopen

PR #258, opened 2021-10-16
forEach iterates over index in stable order

willscott:


v2/index/mhindexsorted_test.go
```diff
@@ -46,6 +46,33 @@ func TestMultiWidthCodedIndex_MarshalUnmarshal(t *testing.T) {
 	requireContainsAll(t, umSubject, records)
 }
 
+func TestMultiWidthCodedIndex_StableIterate(t *testing.T) {
+	rng := rand.New(rand.NewSource(1414))
+	records := generateIndexRecords(t, multihash.SHA2_256, rng)
```
mvdan, line 51: Did this test fail before your fixes? I think the two sources of non-determinism were the maps for "multihash algorithm" and "digest length", so I think you should use two concatenated lists of records - so that each list is different on both axes.

mvdan (approved):
Fix and docs SGTM :) Just one nit to make the test more robust.

v2/index/mhindexsorted_test.go
mvdan: (Otherwise I think your test only ends up with one entry in each of the two layers of maps)

PR #263, opened 2021-11-04
support extraction of unixfs content stored in car files

willscott:
* [x] support symlinks
* [x] round-trip (create then extract) tests


cmd/car/extract.go
```diff
@@ -0,0 +1,192 @@
...
+	cidlink "github.com/ipld/go-ipld-prime/linking/cid"
+	basicnode "github.com/ipld/go-ipld-prime/node/basic"
+	"github.com/urfave/cli/v2"
+)
+
+// ExtractCar pulls files and directories out of a car
+func ExtractCar(c *cli.Context) error {
+	if !c.IsSet("file") {
+		return fmt.Errorf("a file source must be specified")
+	}
```
masih, line 26: Set `Required: true` on flag instead?

Ditto for other similar checks.

cmd/car/extract.go
```diff
@@ -0,0 +1,192 @@
...
+	outputDir, err := os.Getwd()
+	if err != nil {
+		return err
+	}
+	if c.Args().Len() > 0 {
```
masih, line 31:
```suggestion
	if c.Args().Present() {
```

Ditto in other places.

cmd/car/extract.go
```diff
@@ -0,0 +1,192 @@
...
+		outputDir = c.Args().First()
+	}
+
+	bs, err := blockstore.OpenReadOnly(c.String("file"))
```
masih, line 35: Would it make sense to [clean](https://pkg.go.dev/path/filepath#Clean) the path before using?

cmd/car/extract.go
```diff
@@ -0,0 +1,192 @@
...
+		}
+	}
+
+	return nil
+}
+
+func extractRoot(c *cli.Context, ls *ipld.LinkSystem, root cid.Cid, outputDir string) error {
+	if root.Prefix().Codec == cid.Raw {
+		if c.IsSet("verbose") {
+			fmt.Fprintf(os.Stderr, "skipping raw root %s\n", root)
```
masih, line 71: side-note: `c.App.ErrWriter` instead of `os.Stderr`. 

That's technically the right Writer to use. 
There is also one there  for standard output.
OS writers are used directly across the codebase. Please feel free leave TODOs etc. for future improvement.

cmd/car/extract.go
```diff
@@ -0,0 +1,192 @@
...
+	if err := extractDir(c, ls, ufn, outputDir); err != nil {
+		return fmt.Errorf("%s: %w", root, err)
+	}
+
+	return nil
+}
+
+func extractDir(c *cli.Context, ls *ipld.LinkSystem, n ipld.Node, outputDir string) error {
+	// make the directory.
+	os.MkdirAll(outputDir, 0755)
```
masih, line 96: blocker: Should we handle error here?

masih (changes requested):
Small blocker on error handling when making directory otherwise LGTM.
Left suggestions.

rvagg:
Naively writing out files with `path.Join(outputDir, key)` where `key` is just the field you find in a map seems like it might be inviting trouble. Do we do any checking in go-ipfs for whether this file path lands outside of the `outputDir` that we could copy across here? Even allowing `..` or other directory altering characters is going to be a problem here and I'd be surprised if we don't already have code that checks for this in go-ipfs somewhere when creating files.

cmd/car/extract.go
willscott: it should be fine to open and let the OS clean implicitly, right?

willscott:
Okay. There's better accounting to not get in trouble with symlinks or paths outside of the chosen directory now.

cmd/car/extract.go
```diff
@@ -0,0 +1,226 @@
...
+	return joined, nil
+}
+
+func extractDir(c *cli.Context, ls *ipld.LinkSystem, n ipld.Node, outputRoot, outputPath string) error {
+	dirPath, err := resolvePath(outputRoot, outputPath)
+	if err != nil {
+		return err
+	}
+	// make the directory.
+	os.MkdirAll(dirPath, 0755)
```
masih, line 125: `err` check?

cmd/car/extract.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+
+	bs, err := blockstore.OpenReadOnly(c.String("file"))
+	if err != nil {
+		return err
+	}
+
+	ls := cidlink.DefaultLinkSystem()
+	ls.TrustedStorage = true
+	ls.StorageReadOpener = func(_ ipld.LinkContext, l ipld.Link) (io.Reader, error) {
```
hannahhoward, line 40: graphsync's storeutil.LinkSystemForBlockstore does this quite pleasantly -- sets trusted storage, adds a good storage read opener, even includes some optimizations for bytes.Buffer. We should really, really consider putting this in ipld-prime, or somewhere that's obviously accessable.

cmd/car/car.go
```diff
@@ -34,6 +34,26 @@ func main1() int {
 				Usage:  "Detach an index to a detached file",
 				Action: DetachCar,
 			},
+			{
+				Name:    "extract",
+				Aliases: []string{"x"},
+				Usage:   "Extract the contents of a car",
```
hannahhoward, line 40: Looking at the code, it looks like if you give anything other than a car where the roots are DagPB/Raw, it will error. Perhaps it is worth putting a comment like this ("Extracts the contents of a car when the car encodes UnixFS data")

cmd/car/extract.go
```diff
@@ -0,0 +1,226 @@
...
+		return nil
+	}
+
+	pbn, err := ls.Load(ipld.LinkContext{}, cidlink.Link{Cid: root}, dagpb.Type.PBNode)
+	if err != nil {
+		return err
+	}
+	pbnode := pbn.(dagpb.PBNode)
+
+	ufn, err := unixfsnode.Reify(ipld.LinkContext{}, pbnode, ls)
```
hannahhoward, line 80: Not using a the Reify feature of LinkSystem for some reason?

cmd/car/extract.go
```diff
@@ -0,0 +1,226 @@
...
+
+func extractRoot(c *cli.Context, ls *ipld.LinkSystem, root cid.Cid, outputDir string) error {
+	if root.Prefix().Codec == cid.Raw {
+		if c.IsSet("verbose") {
+			fmt.Fprintf(c.App.ErrWriter, "skipping raw root %s\n", root)
+		}
+		return nil
+	}
+
+	pbn, err := ls.Load(ipld.LinkContext{}, cidlink.Link{Cid: root}, dagpb.Type.PBNode)
```
hannahhoward, line 74: This code seems to assume that if you don't have a raw node, you have a root that is a UnixFS Directory, not a regular file (even though extractFile exists as a function). Is this the intent?

hannahhoward (commented):
I am mostly curious if you intended for this code not to support UnixFS files as roots? It appears that way unless I'm missing something.

cmd/car/extract.go
warpfork: There's a glue module for this coming in https://github.com/ipld/go-ipld-prime/pull/279 (which I haven't merged yet, but don't actually know what I'm waiting for, either), so we could use that shortly.

cmd/car/extract.go
willscott: because in `extractDir` we look at the dagpb level data in understanding how and when to do the reification

cmd/car/extract.go
willscott: I believe this is generally the case. if i export a single file it'll be implicitly in a directory in order to provide the file name of the file.

warpfork: fwiw, to compare to other unix stuff, `tar` commands do allow single file no-dir contents, iirc.  Rare and a bit arcane, but possible.

cmd/car/extract.go
willscott: that would be a new coordination with unixfsnode somehow that seems harder to reason about though. I think this is a reasonable setup for applying the reification explicitly when needed.

cmd/car/car.go
willscott: We should think about other conventions that we should try to interpret, like hamts, structured data, etc.

cmd/car/extract.go
willscott: okay for that to go in a follow-up?

rvagg (approved):
this seems to address the concern I had about root escaping, so all paths should strictly only come under the requested output directory

masih (approved):

PR #267, opened 2021-11-16
Seek to start before index generation in `ReadOnly` blockstore

### LINKED ISSUE #265: blockstore.NewReadOnly breaks with io.Reader and CARv1
in case of CARv1 data, `blockstore.NewReadOnly` casts the reader into io.Reader multiple times, and uses that to read versions. This results in
```
error reading car header: invalid header: malformed stream: invalid appearance of uint token; expected start of map
```

To reproduce, get this CARv1 hexdecoded into bytes.Reader, and try to open the car blockstore:
```
38a265726f6f747381d82a5823001220da8ca6ffee4c4cab3d2d5aea4c3320a1100f593c03ce9061f6dbc9bb6de3f36b6776657273696f6e01c1011220da8ca6ffee4c4cab3d2d5aea4c3320a1100f593c03ce9061f6dbc9bb6de3f36b122b0a2212204ccb05f719de64d9754e6d044a05c65f6043b872f8fde2fadaf68d80c9b4442e120018aed4e015122b0a2212209a4038931ac6e4e5794dc1efa1005c14048bfd4e533ff4ad499d172bea53ed74120018aed4e015122b0a221220bd4cb50acf4908a46cc11de1591cd567d2a00aedacad744757f969ed1aeb31a81200188ce5fb020a16080218b3d9bb2e208080e015208080e01520b3d9fb02
```

(just wrapping the `bytes.Reader` into a struct which only provider a ReaderAt works around the bug)

masih:
The blockstore reads the version of the given CAR payload to determine
whether to generate an index for the given payload or not. When the
payload represents a CARv1 and no index is specified, the backing reader
is passed on for index generation. But the version is already read from
the stream. Seek to the beginning of the backing reader before
generating the index so that the index generation mechanism receives the
complete CARv1.

Fixes #265

### DIFF
--- a/v2/blockstore/readonly_test.go
+++ b/v2/blockstore/readonly_test.go
@@ -1,8 +1,10 @@
 package blockstore
 
 import (
+	"bytes"
 	"context"
 	"io"
+	"io/ioutil"
 	"os"
 	"testing"
 	"time"
@@ -268,4 +270,30 @@ func TestReadOnlyErrorAfterClose(t *testing.T) {
 
 	// TODO: test that closing blocks if an AllKeysChan operation is
 	// in progress.
+}
+
+func TestNewReadOnly_CarV1WithoutIndexWorksAsExpected(t *testing.T) {
+	carV1Bytes, err := ioutil.ReadFile("../testdata/sample-v1.car")
+	require.NoError(t, err)
+
+	reader := bytes.NewReader(carV1Bytes)
+	v1r, err := carv1.NewCarReader(reader)
+	require.NoError(t, err)
+	require.Equal(t, uint64(1), v1r.Header.Version)
+
+	// Pick the first block in CARv1 as candidate to check `Get` works.
+	wantBlock, err := v1r.Next()
+	require.NoError(t, err)
+
+	// Seek back to the begining of the CARv1 payload.
+	_, err = reader.Seek(0, io.SeekStart)
+	require.NoError(t, err)
+
+	subject, err := NewReadOnly(reader, nil, UseWholeCIDs(true))
+	require.NoError(t, err)
+
+	// Require that the block is found via ReadOnly API and contetns are as expected.
+	gotBlock, err := subject.Get(wantBlock.Cid())
+	require.NoError(t, err)
+	require.Equal(t, wantBlock, gotBlock)
 }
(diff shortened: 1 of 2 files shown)


v2/blockstore/readonly_test.go
```diff
@@ -269,3 +271,46 @@ func TestReadOnlyErrorAfterClose(t *testing.T) {
 	// TODO: test that closing blocks if an AllKeysChan operation is
 	// in progress.
 }
+
+func TestNewReadOnly_WorksWithIoReader(t *testing.T) {
```
willscott, line 275: it isn't for io.Reader though, right? an io.Reader won't be able to seek back - does this mean that an open call on an actual `io.Reader` stream will fail?

masih: Good point the test name is confusing; thanks for pointing that out

masih: P.S. `ReadOnly` blockstore only takes `io.ReaderAt`

v2/blockstore/readonly_test.go
```diff
@@ -269,3 +271,46 @@ func TestReadOnlyErrorAfterClose(t *testing.T) {
 	// TODO: test that closing blocks if an AllKeysChan operation is
 	// in progress.
 }
+
+func TestNewReadOnly_CarV1WithoutIndexWorksAsExpected(t *testing.T) {
+	carv1Payload := `38a265726f6f747381d82a5823001220da8ca6ffee4c4cab3d2d5aea4c3320a1100f593c03ce9061f6dbc9bb6de3f36b6776657273696f6e01c1011220da8ca6ffee4c4cab3d2d5aea4c3320a1100f593c03ce9061f6dbc9bb6de3f36b122b0a2212204ccb05f719de64d9754e6d044a05c65f6043b872f8fde2fadaf68d80c9b4442e120018aed4e015122b0a2212209a4038931ac6e4e5794dc1efa1005c14048bfd4e533ff4ad499d172bea53ed74120018aed4e015122b0a221220bd4cb50acf4908a46cc11de1591cd567d2a00aedacad744757f969ed1aeb31a81200188ce5fb020a16080218b3d9bb2e208080e015208080e01520b3d9fb02`
```
mvdan, line 276: I wonder if you could reuse one of our existing sample CAR files in testdata, to avoid adding more "blobs" to the test code.

mvdan (approved):
I also wonder if the test is slightly overkill - presumably, you would only need to open the CARv1 file with a blockstore via a `bytes.NewReader` and `Get` one of the CIDs we know is there - none of that should error, where it does error on master.

PR #268, opened 2021-11-19
feat: plumb through context changes

guseggert:


selectivecar.go
```diff
@@ -149,7 +149,7 @@ func (sc SelectiveCarPrepared) Dump(w io.Writer) error {
 		return fmt.Errorf("failed to write car header: %s", err)
 	}
 	for _, c := range sc.cids {
-		blk, err := sc.store.Get(c)
+		blk, err := sc.store.Get(context.TODO(), c)
```
mvdan, line 152: is this on purpose? if we're breaking APIs, one could make an argument for adding a ctx parameter to Dump

mvdan (commented):
Sounds fine with me - I assume you're also taking care of the users of the root-level v0 module?

Also, just to double-check, did you check if the v2 module needs any changes?

aschmahmann:
> Sounds fine with me - I assume you're also taking care of the users of the root-level v0 module?

We're pulling things up through every dependency of the go-ipfs and hydra-booster binaries. Any other consumers (e.g. lotus, estuary, indexer, ...) will have to fend for themselves.

> Also, just to double-check, did you check if the v2 module needs any changes?

Ah, thanks missed that one

selectivecar.go
aschmahmann: It's a question basically. We're breaking go-datastore which is effecting lots of repos anyway, if you'd like to come along for the ride and add contexts to your interfaces that's 👍. If you'd rather us just use `context.TODO()` in all of the go-datastore calls that's fine too.

aschmahmann: To clarify: we're intentionally breaking go-ipfs-blockstore and the interfaces you are using here are clones of go-ipfs-blockstore which means we should break those too.

However, whether that means we should also break `Dump()` or the go-car specific functions is your call.

mvdan: I see. We should be phasing out uses of go-car v0 over time, so I don't think it's worth breaking its APIs just to add contexts, but I'm happy to review PRs in the future if it's an immediate need for anyone.

rvagg: > We should be phasing out uses of go-car v0 over time

Do you mean the API, or the format, or both?

The format's not going away, we're always (for the foreseeable future) going to want to have the ability to read and create v0 format files, does the v2 API let us do that? My understanding was that the v2 API ties you in to writing v2 format files, even if it can read CARv1 files. So I've continued to lean on the v0 API to do everything CARv1 related, is that wrong?

guseggert:
I've updated v2. And also the CLI, by plumbing through the first commit of this PR to the CLI in a second commit using pseudoversions...is that the preferred approach?

v2/blockstore/readonly_test.go
```diff
@@ -18,13 +18,15 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+var bg = context.Background()
```
mvdan, line 21: please don't use a global context; I know it's a no-op and just used in tests, but I would rather avoid sharing any form of state between tests.

v2/blockstore/readwrite_test.go
```diff
@@ -32,6 +32,7 @@ var (
 	rng                       = rand.New(rand.NewSource(1413))
 	oneTestBlockWithCidV1     = merkledag.NewRawNode([]byte("fish")).Block
 	anotherTestBlockWithCidV0 = blocks.NewBlock([]byte("barreleye"))
+	bg                        = context.Background()
```
mvdan, line 35: ditto

v2/example_test.go
```diff
@@ -61,7 +62,7 @@ func ExampleWrapV1File() {
 	if err != nil {
 		panic(err)
 	}
-	fmt.Println(bs.Get(roots[0]))
+	fmt.Println(bs.Get(context.Background(), roots[0]))
```
mvdan, line 65: for examples, I wonder if we should use `context.TODO()` instead, to highlight to the user that they should probably replace the context with something more useful - or consciously use a background context.

I guess it's not super important here, since the CAR blockstore doesn't yet use the context at all, but it might in the future.

mvdan (approved):

guseggert:
This PR Is currently blocked b/c of a [new change on master](https://github.com/ipld/go-car/commit/c35591a559b5fc60bdd107f164b5b25d56d06378) which adds a dependency on https://github.com/ipld/go-ipld-prime/tree/master/storage/bsadapter which uses the old datastore, causing the build to fail. We'll need to upgrade that and plumb it through first before this can be merged.

mvdan:
FYI @willscott @warpfork

whyrusleeping:
I didnt see this PR, but i redid it here: https://github.com/ipld/go-car/pull/273

and it compiles and the tests pass and stuff

mvdan:
It's not equivalent, we'll still want this one - it deals with the v2 module as well.

PR #269, opened 2021-11-22
Propose traversal-based car creation

willscott:
* [x] Complete the file-based alternative writes with no-precalculated size when a file is provided, and then goes pack to fix up the size of the data in the header at the end.
* [x] include carv1 stream support
* [x] Handle options being set on the traversal (repeated blocks)
* [x] Handle options being set for the car (type of index, etc.)
* [x] Roundtrip test


v2/selective.go
```diff
@@ -0,0 +1,221 @@
...
+	ipld "github.com/ipld/go-ipld-prime"
+	"github.com/ipld/go-ipld-prime/linking"
+	cidlink "github.com/ipld/go-ipld-prime/linking/cid"
+	"github.com/ipld/go-ipld-prime/node/basicnode"
+	"github.com/ipld/go-ipld-prime/traversal"
+	"github.com/ipld/go-ipld-prime/traversal/selector"
+	"github.com/multiformats/go-varint"
+)
+
+// PrepareTraversal walks through the proposed dag traversal to learn it's total size in order to be able to
```
rvagg, line 19: Lotus evolved away from needing this, so I don't think there's a consumer for _prepare+dump_ anymore, it's just _write_ now, thankfully. There used to be a need to get the size up-front to set up the commp calculation with padding, but that's not necessary anymore. So you could decomplicate this by removing this functionality if you like.

willscott: If we want to write out to a stream - e.g. network, this 2x scan would still be more efficient than having to write it out to a file, touch up the header, and then as a second step send the whole thing, i think

rvagg: Oh, so this would be for a CARv2 format and you need the offset and all that? Because for a selector-based CARv1 you have the root already, or else you can't run the traversal.

willscott: yeah, this is to get the data size field correct in the carv2 header when streaming.

I can make a carv1-only stream version that can do it in one pass as well. I suppose that's probably useful as well.

willscott:
I think i have all the code here doing what I want. starting in on tests.

This would be a great time for reviews if you want to request changes to naming

v2/internal/loader/counting_loader.go
```diff
@@ -0,0 +1,67 @@
...
+// link system which trigger block reads, the size of the block as it would
+// appear in a CAR file is added to the counter (included the size of the
+// CID and the varint length for the block data).
+func CountingLinkSystem(ls ipld.LinkSystem) (ipld.LinkSystem, ReadCounter) {
+	c := counter{}
+	return linking.LinkSystem{
+		EncoderChooser:     ls.EncoderChooser,
+		DecoderChooser:     ls.DecoderChooser,
+		HasherChooser:      ls.HasherChooser,
+		StorageWriteOpener: ls.StorageWriteOpener,
```
warpfork, line 49: You can skip a considerable amount of this by relying on pass-by-value to copy them for you.  (And be a bit more future proof in case there are ever any added fields, that way, too.)

v2/internal/loader/writing_loader.go
```diff
@@ -0,0 +1,119 @@
...
+		size:  initialOffset,
+		code:  indexCodec,
+		rcrds: make([]index.Record, 0),
+	}
+	return linking.LinkSystem{
+		EncoderChooser:     ls.EncoderChooser,
+		DecoderChooser:     ls.DecoderChooser,
+		HasherChooser:      ls.HasherChooser,
+		StorageWriteOpener: ls.StorageWriteOpener,
+		StorageReadOpener: func(lc linking.LinkContext, l ipld.Link) (io.Reader, error) {
```
warpfork, line 104: Okay, I'm not gonna try to steer here, this is more a question for me to understand how to serve this area better in the future: Did you go through a decision process on whether to do this wrap `StorageReadOpener` approach vs do some wrappers around the new `storage.*` APIs?

Is this wrapping strategy just familiar already?

Is this wrapping strategy preferable because it seems like the most broadly applicable place to intercept things?

Is this wrapping strategy preferable because it's clear how it composes with other StorageReadOpener magic, such as how graphsync uses it to hook block load events?

Is it because figuring out how to wrap the `storage.*` APIs is daunting because it would've required thinking about all the places feature detection might try kick in?

Other?

(These are not blocking questions; they just could be useful info for understanding the value of any future iterations on storage and link loading APIs and other possible event callbacks around them.)

willscott: I think it's that the object that is expected to be passed around is a `LinkSystem` - that linksystem will already have the `storage.*` object applied to it and turned into a `StorageReadOpener`, and in this library I won't know what the underlying readable storage was to be able to wrap it. If i assume the right thing to take in is a link system, which seems right because i need to care about things like 'should i hash reads', 'what reifiers are present', etc, then this is the place i can intercept block reads

v2/selective.go
```diff
@@ -0,0 +1,268 @@
...
+	"github.com/ipld/go-ipld-prime/traversal/selector"
+)
+
+// ErrSizeMismatch is returned when a written traversal realizes the written header size does not
+// match the actual number of car bytes written.
+var ErrSizeMismatch = fmt.Errorf("car-error-sizemismatch")
+
+// ErrOffsetImpossible is returned when specified paddings or offsets of either a wrapped carv1
+// or index cannot be satisfied based on the data being written.
+var ErrOffsetImpossible = fmt.Errorf("car-error-offsetimpossible")
```
masih, line 28: * Maybe `Invalid` or `Illegal` instead of `Impossible`? They tend to be the terms more commonly used to signal such issues.
* Human readable message 
* `errors.New` instead of `fmt.Errorf` since there isn't anything to format

v2/selective.go
```diff
@@ -0,0 +1,268 @@
...
+	tc.size = uint64(tc.size)
+	if _, err = tc.WriteV2Header(fp); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// V1Traversal walks through the proposed dag traversal and writes a carv1 to the provided io.Writer
+func V1Traversal(ctx context.Context, ls *ipld.LinkSystem, root cid.Cid, selector ipld.Node, writer io.Writer, opts ...Option) (uint64, error) {
```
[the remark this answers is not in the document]
masih: It'd be awesome to add unit tests for this one too.

v2/selective.go
```diff
@@ -0,0 +1,268 @@
...
+		selector: selector,
+		ls:       ls,
+		opts:     opts,
+	}
+
+	len, _, err := tc.WriteV1(writer)
+	return len, err
+}
+
+// Writer is an interface allowing writing a car with the data specified by a selector.
```
masih, line 117: I am curious what in this interface is selector-specific? 
* If it is meant to be selects-specific I'd recommend qualifying  that in the name since it's in the root of CARv2 library in case we add writers in the future.

v2/selective.go
```diff
@@ -0,0 +1,268 @@
...
+type Writer interface {
+	io.WriterTo
+}
+
+type traversalCar struct {
```
masih, line 122:
```suggestion
var _ Writer = (*traversalCar)(nil)

type traversalCar struct {
```

v2/selective.go
```diff
@@ -0,0 +1,268 @@
...
+	size     uint64
+	ctx      context.Context
+	root     cid.Cid
+	selector ipld.Node
+	ls       *ipld.LinkSystem
+	opts     []Option
```
masih, line 128: Would it make sense to use `Options` struct type and avoid repeated `ApplyOptions` across receiver functions of `traveralCar`? 
That should keep the exported functions as they are, i.e. taking varargs `[]Option`, and apply options once when instantiating a new `traveraalCar`.

v2/selective.go
```diff
@@ -0,0 +1,268 @@
...
+// execution when building a SelectiveCar.
+func MaxTraversalLinks(MaxTraversalLinks uint64) Option {
+	return func(sco *Options) {
+		sco.MaxTraversalLinks = MaxTraversalLinks
+	}
+}
+
+// PrepareTraversal walks through the proposed dag traversal to learn it's total size in order to be able to
+// stream out a car to a writer in the expected traversal order in one go.
+func PrepareTraversal(ctx context.Context, ls *ipld.LinkSystem, root cid.Cid, selector ipld.Node, opts ...Option) (Writer, error) {
```
masih, line 43: `NewTraversalWriter`? or maybe rename the term `Traversal` to `Selective` across this file? That'd make it consistent with the file name and `NewSelectiveWriter` to me sounds closer to what this function does.

v2/selective.go
```diff
@@ -0,0 +1,268 @@
...
+	h := NewHeader(tc.size)
+	conf := ApplyOptions(tc.opts...)
+	if p := conf.DataPadding; p > 0 {
+		h = h.WithDataPadding(p)
+	}
+	if p := conf.IndexPadding; p > 0 {
+		h = h.WithIndexPadding(p)
+	}
+	// TODO: support calculation / inclusion of the index.
+	h.IndexOffset = 0
```
masih, line 179: I am curious what is the work that's left to do here? If we know the payload size and paddings, which I think we do, We can set this value; right? `carv2.Header` `With*` functions should already update index offset accordingly.

If there is a body of work that needs to be done here before we can set the offset, then would it make sense to panic if any index codec other than "no-index" is given to make it obvious there is a TODO?
Right now it seems the index is written if codec allows it to but index offset is never set.

v2/selective.go
```diff
@@ -0,0 +1,268 @@
...
+	hn, err := h.WriteTo(w)
+	if err != nil {
+		return int64(n) + hn, err
+	}
+	hn += int64(n)
+
+	// We include the initial data padding after the carv2 header
+	if h.DataOffset > uint64(hn) {
+		buf := make([]byte, h.DataOffset-uint64(hn))
+		n, err = w.Write(buf)
```
masih, line 189: This could be less than optimum in terms of memory consumption if the padding is large.
Use buffered write instead?

v2/selective.go
```diff
@@ -0,0 +1,268 @@
...
+	return n, err
+}
+
+func (tc *traversalCar) WriteV2Header(w io.Writer) (int64, error) {
+	n, err := w.Write(Pragma)
+	if err != nil {
+		return int64(n), err
+	}
+
+	h := NewHeader(tc.size)
```
masih, line 170: Thinking out loud, in a separate PR we probably should change `NewHeader` to take options and encapsulate the gymnastics needed to set correct header values.

v2/index/index.go
```diff
@@ -40,7 +43,7 @@ type (
 		Codec() multicodec.Code
 
 		// Marshal encodes the index in serial form.
-		Marshal(w io.Writer) error
+		Marshal(w io.Writer) (uint64, error)
```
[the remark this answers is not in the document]
willscott: Is anything using this index package directly where this signature change would break them?

v2/internal/loader/counting_loader.go
```diff
@@ -0,0 +1,61 @@
...
+	clc := ls
+	clc.StorageReadOpener = func(lc linking.LinkContext, l ipld.Link) (io.Reader, error) {
+		r, err := ls.StorageReadOpener(lc, l)
+		if err != nil {
+			return nil, err
+		}
+		buf := bytes.NewBuffer(nil)
+		n, err := buf.ReadFrom(r)
+		if err != nil {
+			return nil, err
```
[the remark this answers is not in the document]
willscott: none of the block will be available to be added to the output stream, so in terms of the expected car output stream length none of those read bytes would be counted.

v2/index/index.go
```diff
@@ -15,6 +15,9 @@ import (
 	"github.com/ipfs/go-cid"
 )
 
+// CarIndexNoIndex is a sentinal value used as a multicodec code for the index indicating no index.
+const CarIndexNoIndex = 0x300000
```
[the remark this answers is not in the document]
willscott: this is within the designated internal range that can't be put in the CSV

v2/selective.go
```diff
@@ -0,0 +1,268 @@
...
+	"github.com/ipld/go-ipld-prime/linking"
+	cidlink "github.com/ipld/go-ipld-prime/linking/cid"
+	"github.com/ipld/go-ipld-prime/node/basicnode"
+	"github.com/ipld/go-ipld-prime/traversal"
+	"github.com/ipld/go-ipld-prime/traversal/selector"
+)
+
+// ErrSizeMismatch is returned when a written traversal realizes the written header size does not
+// match the actual number of car bytes written.
+var ErrSizeMismatch = fmt.Errorf("car-error-sizemismatch")
```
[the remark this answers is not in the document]
willscott: this was trying to follow the code-usable error semantics proposed at https://github.com/warpfork/go-ree

v2/selective.go
willscott: added todo.

v2/index/index.go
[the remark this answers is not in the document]
masih: I think the answer is we don't know. Would be great if we don't break the API if we don't have to and use standard interfaces, specially when the signature matches exactly

v2/selective.go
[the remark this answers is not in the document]
masih: Where is the rest of [schema](https://github.com/warpfork/go-ree#schema) fields? As it stands code alone makes the errors less human readable; no?

v2/index/index.go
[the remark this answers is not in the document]
willscott: This is the trap IPFS fell into. the answer can't be 'because we don't know we are unwilling to deprecate any API'. I think we stick with the smaller API that we want, rather than keeping all previous versions in 'deprecated'

v2/selective.go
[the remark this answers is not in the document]
willscott: everything else is optional. This dashed format allows for partial matching, so e.g. allows an upper app to trap all 'car' derived errors.

masih (approved):

v2/index/index.go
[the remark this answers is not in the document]
masih: I need to learn more about what happened in IPFS. In this specific case I do think it makes sense to reuse an existing interface used already elsewhere. From API design perspective at the very least. 

This is not a blocker for the purposes of this PR review from my perspective.

PR #281, opened 2022-01-06
make specification of root cid in get-dag command optional

willscott:
add a test script


cmd/car/testdata/script/get.txt
```diff
@@ -0,0 +1,6 @@
+env SAMPLE_CID='bafy2bzaceaycv7jhaegckatnncu5yugzkrnzeqsppzegufr35lroxxnsnpspu'
```
mvdan, line 1: following the other files, shouldn't we call this get-dag.txt?

if you prefer get.txt, then I'd join get-block.txt here.

mvdan (approved):

cmd/car/testdata/script/get.txt
willscott: the go file under test is `get.go` which has both `get-block` and `get-dag`. what do you think is more reasonable?

mvdan: If you expect the test scripts to grow further, I'd keep them separate

Issue #282, opened 2022-01-13
carv2.NewBlockReader cannot be used with a stream


willscott:
`NewBlockReader` takes an `io.Reader`, however when passed `os.Stdin` with a carv2 payload, the call to new will error with "seek /dev/stdin: illegal seek"

It's unclear how this is trying to seek on an `io.Reader` in the first place.

mvdan:
Oops, that sounds like at least an oversight.

PR #283, opened 2022-01-13
add `car root` command

willscott:
allow listing out specified roots in a car file


cmd/car/testdata/script/root.txt
```diff
@@ -0,0 +1,15 @@
+car root ${INPUTS}/sample-v1.car
+cmp stdout v1root.txt
+
+car root ${INPUTS}/sample-wrapped-v2.car
+cmp stdout v2root.txt
+
+stop stdin test needs car fix
```
willscott, line 7: note that the stdin variant of the command currently has issues. I want to get this command available in file mode today for users, so that work is filed as #282

mvdan: link the issue as a TODO?

cmd/car/root.go
```diff
@@ -0,0 +1,39 @@
...
+
+	carv2 "github.com/ipld/go-car/v2"
+	"github.com/ipld/go-car/v2/blockstore"
+	"github.com/urfave/cli/v2"
+)
+
+// CarRoot prints the root CID in a car
+func CarRoot(c *cli.Context) (err error) {
+	if c.Args().Len() >= 1 {
+		bs, err := blockstore.OpenReadOnly(c.Args().First())
```
mvdan, line 15: I think it would be nicer to use NewBlockReader for both; it does less work than opening a blockstore, as that often does extras we don't need like loading the index into memory. The only difference is that you need to do an `os.Open` when given a filename.

mvdan (approved):

cmd/car/testdata/script/root.txt
mvdan: also, FYI, I usually do:

```
stop # explain the reason
```

same with skip etc

PR #284, opened 2022-01-13
Allow extracton of a raw unixfs file

willscott:
If the root of a car is a unixfs file, without the containing directory entry, we can still extract it, we just don't have a filename for it.

### DIFF
--- a/cmd/car/extract.go
+++ b/cmd/car/extract.go
@@ -108,7 +108,7 @@ func extractRoot(c *cli.Context, ls *ipld.LinkSystem, root cid.Cid, outputDir st
 				return err
 			}
 			if ufsNode.DataType.Int() == data.Data_File || ufsNode.DataType.Int() == data.Data_Raw {
-				if err := extractFile(c, ls, pbnode, path.Join(outputResolvedDir, "unknown")); err != nil {
+				if err := extractFile(c, ls, pbnode, filepath.Join(outputResolvedDir, "unknown")); err != nil {
 					return err
 				}
 			}
--- a/cmd/car/extract.go
+++ b/cmd/car/extract.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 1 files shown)


cmd/car/extract.go
```diff
@@ -92,6 +94,26 @@ func extractRoot(c *cli.Context, ls *ipld.LinkSystem, root cid.Cid, outputDir st
 		}
 	}
 	if err := extractDir(c, ls, ufn, outputResolvedDir, "/"); err != nil {
+		if err == ErrNotDir {
```
mvdan, line 97: nit: use `errors.Is`

mvdan: also, you could unindent the not-dir handling code like:

```
if !errors.Is(err, ErrNotDir) {
    return fmt.Errorf(...)
}
// non-dir code lines here
```

cmd/car/extract.go
```diff
@@ -92,6 +94,26 @@ func extractRoot(c *cli.Context, ls *ipld.LinkSystem, root cid.Cid, outputDir st
...
+			}
+			ufsBytes, err := ufsData.AsBytes()
+			if err != nil {
+				return err
+			}
+			ufsNode, err := data.DecodeUnixFSData(ufsBytes)
+			if err != nil {
+				return err
+			}
+			if ufsNode.DataType.Int() == data.Data_File || ufsNode.DataType.Int() == data.Data_Raw {
```
mvdan, line 110: nit: would be a bit shorter if you use a switch, as you deduplicate the left side

cmd/car/extract.go
```diff
@@ -92,6 +94,26 @@ func extractRoot(c *cli.Context, ls *ipld.LinkSystem, root cid.Cid, outputDir st
...
+				if err := extractFile(c, ls, pbnode, path.Join(outputResolvedDir, "unknown")); err != nil {
```
mvdan, line 111:
```suggestion
				if err := extractFile(c, ls, pbnode, filepath.Join(outputResolvedDir, "unknown")); err != nil {
```

mvdan (approved):

PR #287, opened 2022-01-20
add `car detach-index list` to list detached index contents

### LINKED ISSUE #286: car cli should provide a way to list / search detached indexes
we have cases where there are detached indexes floating around. it would be useful to check membership and to enumerate their contents

willscott:
fix #286

### DIFF
--- a/cmd/car/detach.go
+++ b/cmd/car/detach.go
@@ -6,6 +6,8 @@ import (
 	"os"
 
 	carv2 "github.com/ipld/go-car/v2"
+	"github.com/ipld/go-car/v2/index"
+	"github.com/multiformats/go-multihash"
 	"github.com/urfave/cli/v2"
 )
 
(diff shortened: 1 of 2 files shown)


cmd/car/detach.go
```diff
@@ -33,3 +35,35 @@ func DetachCar(c *cli.Context) error {
...
+			fmt.Printf("%s %d\n", mh, offset)
+			return nil
+		})
+		if err != nil {
+			return err
+		}
+		return nil
+	}
+
+	return fmt.Errorf("index of codec %d is not iterable", idx.Codec())
```
masih, line 68: Would it be useful to print the codec name here instead?

```suggestion
	return fmt.Errorf("index of codec %s is not iterable", idx.Codec())
```

masih (approved):

cmd/car/detach.go
willscott: can i just do `%T` to get a decent type name?

masih: I _think_ that'd print `multicodec.Code`.

PR #288, opened 2022-01-20
feat: add option to create blockstore that writes a plain CARv1

rvagg:
Use-case: I _only_ want a CARv1 output but I want to use the blockstore interface. I don't want to have to extract the CARv1 output from a CARv2 afterward.

Is this a reasonable way to achieve this end?

It would mean that the `car create ...` CLI command could have a `--version` flag which only needs to pass this new option to the blockstore.


willscott (approved):
🎉

v2/blockstore/readwrite.go
```diff
@@ -158,36 +176,39 @@ func (b *ReadWrite) resumeWithRoots(roots []cid.Cid) error {
 		// Or the write must have failed before pragma was written.
 		return err
 	}
-	if version != 2 {
-		// The file is not a CARv2 and we cannot resume from it.
+	if !((version == 1 && !v2) || (version == 2 && v2)) {
```
mvdan, line 179: I'm having a hard time parsing this in my head. How about:

```
switch {
case version == 1 && !v2:
case version == 2 && v2:
default:
    return fmt.Errorf(...)
}
```

v2/blockstore/readwrite.go
```diff
@@ -354,6 +378,11 @@ func (b *ReadWrite) Discard() {
 // for more efficient subsequent read.
 // After this call, the blockstore can no longer be used.
 func (b *ReadWrite) Finalize() error {
+	if b.opts.WriteAsCarV1 {
+		b.Discard()
```
mvdan, line 382: I'd document here why we have nothing to do in the v1 case. For example, right now I forget why Finalize doesn't need to write the carv1 data at all - is it already written to disk?

mvdan (approved):

v2/blockstore/readwrite.go
rvagg: I'll have to take you on your word that that's more readable to Go eyes than the original

v2/blockstore/readwrite.go
rvagg: Yeah, already written to disk, no trailer needed for a v1. I'm also changing to a direct `b.ronly.Close()` call instead of `b.Discard()` to be less confusing.

mvdan (approved):

PR #289, opened 2022-01-20
feat: --version selector for car create & update deps for cmd/car

rvagg:
Builds on #288, review can probably wait until (if) that PR is merged.


cmd/car/create.go
```diff
@@ -43,7 +43,14 @@ func CreateCar(c *cli.Context) error {
 	}
 	proxyRoot := cid.NewCidV1(uint64(multicodec.DagPb), hash)
 
-	cdest, err := blockstore.OpenReadWrite(c.String("file"), []cid.Cid{proxyRoot})
+	options := []car.Option{}
+	if c.Int("version") == 1 {
+		options = []car.Option{blockstore.WriteAsCarV1(true)}
+	} else if c.Int("version") != 2 {
+		return fmt.Errorf("invalid CAR version %d", c.Int("version"))
+	}
```
mvdan, line 51:
```suggestion
	switch c.Int("version") {
	case 1:
		options = []car.Option{blockstore.WriteAsCarV1(true)}
	case 2:
		// already the default
	default:
		return fmt.Errorf("invalid CAR version %d", c.Int("version"))
	}
```

cmd/go.mod
```diff
@@ -6,15 +6,17 @@ require (
...
-	github.com/multiformats/go-multihash v0.0.17-0.20211012170226-654b06d7912a
+	github.com/ipld/go-ipld-prime v0.14.4
+	github.com/multiformats/go-multicodec v0.4.0
+	github.com/multiformats/go-multihash v0.1.0
 	github.com/multiformats/go-varint v0.0.6
 	github.com/rogpeppe/go-internal v1.8.1-0.20210923151022-86f73c517451
 	github.com/urfave/cli/v2 v2.3.0
 )
+
+replace github.com/ipld/go-car/v2 => ../v2
```
mvdan, line 22: remember to undo this when you've rebased the PR

mvdan (approved):
cmd/car changes SGTM.

I initially wrote a comment about the CLI flag being an int whereas the Go option is a boolean, saying that they should use the same type. But it makes sense; the carv2 Go library will only ever support v1+v2, and a future v3 will most likely be as part of a new Go module major version. Whereas the cmd/car CLI is fairly high level and basic, so it will most likely be able to add v3 support without breaking backwards compatibility, meaning that an integer is best.

mvdan (approved):
:ok_hand:

PR #291, opened 2022-01-26
Add a 'skip' parameter to writev1 so that the beginning of a car can …

willscott:
…be skipped

### DIFF
--- a/v2/internal/loader/writing_loader.go
+++ b/v2/internal/loader/writing_loader.go
@@ -13,10 +13,11 @@ import (
 )
 
 type writerOutput struct {
-	w     io.Writer
-	size  uint64
-	code  multicodec.Code
-	rcrds []index.Record
+	w      io.Writer
+	size   uint64
+	toSkip uint64
+	code   multicodec.Code
+	rcrds  []index.Record
 }
 
 func (w *writerOutput) Size() uint64 {
@@ -50,17 +51,31 @@ type writingReader struct {
 }
 
 func (w *writingReader) Read(p []byte) (int, error) {
+	buf := bytes.NewBuffer(nil)
 	if w.wo != nil {
 		// write the cid
 		size := varint.ToUvarint(uint64(w.len) + uint64(len(w.cid)))
-		if _, err := w.wo.w.Write(size); err != nil {
+		if _, err := buf.Write(size); err != nil {
 			return 0, err
 		}
-		if _, err := w.wo.w.Write([]byte(w.cid)); err != nil {
+		if _, err := buf.Write([]byte(w.cid)); err != nil {
 			return 0, err
 		}
 		cpy := bytes.NewBuffer(w.r.(*bytes.Buffer).Bytes())
-		if _, err := cpy.WriteTo(w.wo.w); err != nil {
+		if _, err := cpy.WriteTo(buf); err != nil {
+			return 0, err
+		}
+		out := buf.Bytes()
+		if w.wo.toSkip > 0 {
+			if w.wo.toSkip >= uint64(len(out)) {
+				w.wo.toSkip -= uint64(len(out))
+				out = []byte{}
+			} else {
+				out = out[w.wo.toSkip:]
+				w.wo.toSkip = 0
+			}
+		}
+		if _, err := bytes.NewBuffer(out).WriteTo(w.wo.w); err != nil {
 			return 0, err
 		}
 
@@ -89,12 +104,13 @@ func (w *writingReader) Read(p []byte) (int, error) {
 // The `initialOffset` is used to calculate the offsets recorded for the index, and will be
 //   included in the `.Size()` of the IndexTracker.
 // An indexCodec of `index.CarIndexNoIndex` can be used to not track these offsets.
-func TeeingLinkSystem(ls ipld.LinkSystem, w io.Writer, initialOffset uint64, indexCodec multicodec.Code) (ipld.LinkSystem, IndexTracker) {
+func TeeingLinkSystem(ls ipld.LinkSystem, w io.Writer, initialOffset uint64, skip uint64, indexCodec multicodec.Code) (ipld.LinkSystem, IndexTracker) {
 	wo := writerOutput{
-		w:     w,
-		size:  initialOffset,
-		code:  indexCodec,
-		rcrds: make([]index.Record, 0),
+		w:      w,
+		size:   initialOffset,
+		toSkip: skip,
+		code:   indexCodec,
+		rcrds:  make([]index.Record, 0),
 	}
 
 	tls := ls
--- a/v2/selective.go
+++ b/v2/selective.go
@@ -1,6 +1,7 @@
 package car
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"io"
@@ -110,7 +111,7 @@ func TraverseV1(ctx context.Context, ls *ipld.LinkSystem, root cid.Cid, selector
 		opts:     ApplyOptions(opts...),
 	}
 
-	len, _, err := tc.WriteV1(writer)
+	len, _, err := tc.WriteV1(0, writer)
 	return len, err
 }
 
@@ -135,7 +136,7 @@ func (tc *traversalCar) WriteTo(w io.Writer) (int64, error) {
 	if err != nil {
 		return n, err
 	}
-	v1s, idx, err := tc.WriteV1(w)
+	v1s, idx, err := tc.WriteV1(0, w)
 	n += int64(v1s)
 
 	if err != nil {
@@ -200,21 +201,35 @@ func (tc *traversalCar) WriteV2Header(w io.Writer) (int64, error) {
 	return hn, nil
 }
 
-func (tc *traversalCar) WriteV1(w io.Writer) (uint64, index.Index, error) {
+// WriteV1 writes a v1 car to the writer, w, except for the first `skip` bytes.
+// Returns bytes written, an index of what was written, or error if one occured.
+func (tc *traversalCar) WriteV1(skip uint64, w io.Writer) (uint64, index.Index, error) {
+	written := uint64(0)
+
 	// write the v1 header
 	c1h := carv1.CarHeader{Roots: []cid.Cid{tc.root}, Version: 1}
-	if err := carv1.WriteHeader(&c1h, w); err != nil {
-		return 0, nil, err
-	}
 	v1Size, err := carv1.HeaderSize(&c1h)
 	if err != nil {
-		return v1Size, nil, err
+		return 0, nil, err
+	}
+	if skip < v1Size {
+		buf := bytes.NewBuffer(nil)
+		if err := carv1.WriteHeader(&c1h, buf); err != nil {
+			return 0, nil, err
+		}
+		if _, err := w.Write(buf.Bytes()[skip:]); err != nil {
+			return 0, nil, err
+		}
+		written = v1Size - skip
+		skip = 0
+	} else {
+		skip -= v1Size
 	}
 
 	// write the block.
-	wls, writer := loader.TeeingLinkSystem(*tc.ls, w, v1Size, tc.opts.IndexCodec)
+	wls, writer := loader.TeeingLinkSystem(*tc.ls, w, v1Size, skip, tc.opts.IndexCodec)
 	err = traverse(tc.ctx, &wls, tc.root, tc.selector, tc.opts)
-	v1Size = writer.Size()
+	v1Size = writer.Size() - v1Size + written
 	if err != nil {
 		return v1Size, nil, err
 	}


v2/internal/loader/writing_loader.go
```diff
@@ -50,17 +51,31 @@ type writingReader struct {
 }
 
 func (w *writingReader) Read(p []byte) (int, error) {
+	buf := bytes.NewBuffer(nil)
```
dirkmc, line 54: Is it possible to avoid creating a new buffer here unless the skip falls within this block?

willscott: the complexity of getting the potential partial-write edge cases of each of the 3 internal writes worries me a little bit, but yeah i'll take a look

v2/selective.go
```diff
@@ -38,17 +40,36 @@ func MaxTraversalLinks(MaxTraversalLinks uint64) Option {
 	}
 }
 
+// WithV1Size sets the expected v1 size of the car being written if it is known in advance.
+func WithV1Size(size uint64) Option {
```
masih, line 44:
```suggestion
func WithDataPayloadSize(size uint64) Option {
```
The term `DataPayload` is used to describe the inner CARv1 in [CARv2 spec](https://ipld.io/specs/transport/car/carv2/#carv1-data-payload). 
It's also a term used across other v2 APIs.

v2/selective.go
```diff
@@ -101,19 +123,40 @@ func TraverseToFile(ctx context.Context, ls *ipld.LinkSystem, root cid.Cid, sele
...
+		opts:     conf,
 	}
 
-	len, _, err := tc.WriteV1(writer)
+	len, _, err := tc.WriteV1(tc.ctx, 0, writer)
 	return len, err
 }
 
+// CreateV1Reader creates an io.ReadSeeker that can be used to copy out the carv1 contents of a car.
+func CreateV1Reader(ctx context.Context, ls *ipld.LinkSystem, root cid.Cid, selector ipld.Node, opts ...Option) (io.ReadSeeker, error) {
```
masih, line 141: * Use `New` instead of `Create` for consistency with existing APIs?
* This is a root level function. Considering it applies to the "selective" read/write context, can I ask you to rename this to something less generic.

willscott: still waiting to figure out what the partial traversal API in ipld-prime will look like. this may be a function on a traversalCar instead, depending on where we want shared cache state between different traversals to live.

willscott:
The remaining potential work here is allow skip to be an arbitrary byte offset in the car file. currently it's limited to block boundaries, and because of the API it's difficult to know which block boundary nearest to the requested skip offset has been used.
If it's acceptable for the caller to skip only to block boundaries, this could be sufficient as-is, otherwise the wrapper for the individual reader needs to be updated to also partially skip.

willscott:
I cleaned up the tee-ing writer and added a test that seems to verify that we can use this to advance to arbitrary byte offsets in a car.

@rvagg maybe can provide a first round of review

v2/selective.go
willscott: I have now lost context here and am not sure if my argument about partial traversal holds

v2/internal/loader/counting_loader.go
```diff
@@ -9,13 +9,13 @@ import (
 	"github.com/multiformats/go-varint"
 )
 
-// counter tracks how much data has been read.
-type counter struct {
-	totalRead uint64
+// Counter tracks how much data has been read.
+type Counter struct {
```
rvagg, line 13: why do we use a struct for this? is it something to do with needing a pointer to it? why isn't a `*uint64` good enough for this purpose?

v2/traversal/resumption.go
```diff
@@ -0,0 +1,257 @@
...
+func newPath(link datamodel.Link, at uint64) *pathNode {
+	return &pathNode{
+		link:     link,
+		offset:   at,
+		children: make(map[datamodel.PathSegment]*pathNode),
+	}
+}
+
+func (pn pathNode) addPath(p []datamodel.PathSegment, link datamodel.Link, at uint64) {
+	if len(p) == 0 {
```
rvagg, line 35: is this just for the root case where we've made a `newPath()` and don't want to add the root to itself?

v2/traversal/resumption.go
```diff
@@ -0,0 +1,257 @@
...
+	if nc != nil {
+		return nc.offset, nil
+	}
+
+	return 0, errInvalid
+}
+
+// TraverseResumer allows resuming a progress from a previously encountered path in the selector.
+type TraverseResumer interface {
+	RewindToPath(from datamodel.Path) error
```
rvagg, line 126: am I right in assuming we don't have a use-case for this at the moment, that we're just supporting it because it's not a difficult extension of doing it to an offset?

got thoughts on what a use-case for might me?

v2/internal/loader/writing_loader.go
```diff
@@ -46,25 +49,50 @@ type IndexTracker interface {
...
 
+var _ IndexTracker = (*indexingWriter)(nil)
+
 type writingReader struct {
 	r   io.Reader
-	len int64
+	buf []byte
 	cid string
-	wo  *writerOutput
+	wo  *indexingWriter
```
rvagg, line 58: maybe rename this so it's not got an unexplainable name for future eyes?

```suggestion
	iw  *indexingWriter
```

v2/internal/loader/writing_loader.go
```diff
@@ -46,25 +49,50 @@ type IndexTracker interface {
...
+		sizeBytes := varint.ToUvarint(uint64(n) + uint64(len(w.cid)))
+		writeBuf := buf.Bytes()[varint.MaxLenUvarint63-len(sizeBytes):]
+		w.buf = buf.Bytes()[varint.MaxLenUvarint63+len(w.cid):]
+		_ = copy(writeBuf[:], sizeBytes)
+
+		size := len(writeBuf)
+		if w.wo.toSkip > 0 {
+			if w.wo.toSkip >= uint64(len(writeBuf)) {
+				w.wo.toSkip -= uint64(len(writeBuf))
+				writeBuf = []byte{}
```
rvagg, line 88:
```suggestion
				// will cause the WriteTo() below to be a noop, we need to skip this entire block
				writeBuf = []byte{}
```

v2/internal/loader/writing_loader.go
```diff
@@ -46,25 +49,50 @@ type IndexTracker interface {
...
 		}
-		if _, err := w.wo.w.Write([]byte(w.cid)); err != nil {
+		// write the block
+		n, err := io.Copy(buf, w.r)
+		if err != nil {
 			return 0, err
 		}
-		cpy := bytes.NewBuffer(w.r.(*bytes.Buffer).Bytes())
-		if _, err := cpy.WriteTo(w.wo.w); err != nil {
+		sizeBytes := varint.ToUvarint(uint64(n) + uint64(len(w.cid)))
```
rvagg, line 79:
```suggestion
		// write the varint size prefix and trim the unneeded prefix padding we allocated
		sizeBytes := varint.ToUvarint(uint64(n) + uint64(len(w.cid)))
```

v2/selective.go
```diff
@@ -202,21 +250,38 @@ func (tc *traversalCar) WriteV2Header(w io.Writer) (int64, error) {
...
+		if _, err := w.Write(buf.Bytes()[skip:]); err != nil {
+			return 0, nil, err
+		}
+		written = v1Size - skip
+		skip = 0
+	} else {
+		skip -= v1Size
 	}
 
 	// write the block.
```
rvagg, line 278:
```suggestion
	// write the blocks
```

v2/internal/loader/writing_loader.go
```diff
@@ -75,11 +103,19 @@ func (w *writingReader) Read(p []byte) (int, error) {
 			Cid:    c,
 			Offset: w.wo.size,
 		}
-		w.wo.size += uint64(w.len) + uint64(len(size)+len(w.cid))
-
+		w.wo.size += uint64(size)
 		w.wo = nil
 	}
 
+	if w.buf != nil {
```
rvagg, line 110:
```suggestion
	if w.buf != nil {
		// we've already read the block from the parent reader for writing the CAR block section (above),
		// so we need to pass those bytes on in whatever chunk size the caller wants
```

v2/internal/loader/writing_loader.go
```diff
@@ -46,25 +49,50 @@ type IndexTracker interface {
 	Index() (index.Index, error)
 }
 
+var _ IndexTracker = (*indexingWriter)(nil)
+
 type writingReader struct {
```
rvagg, line 54:
```suggestion
// writingReader is used on a per-block basis for the TeeingLinkSystem's StorageReadOpener, we use it
// to intercept block reads and construct CAR section output for that block, passing that section data to
// indexingWriter, while also passing the plain binary block data back to the LinkSystem caller (which
// we expect to be a traversal operation).
// Additionally, if we are performing a "skip" of initial bytes for this CAR, we track the byte count as we
// construct the CAR section data and decide when and how much to write out to the indexingWriter.
// Skip operations don't impact the downstream LinkSystem user (traversal), but they do impact what's
// written out via the indexingWriter.
type writingReader struct {
```

v2/internal/loader/writing_loader.go
```diff
@@ -46,25 +49,50 @@ type IndexTracker interface {
...
 		}
-		cpy := bytes.NewBuffer(w.r.(*bytes.Buffer).Bytes())
-		if _, err := cpy.WriteTo(w.wo.w); err != nil {
+		sizeBytes := varint.ToUvarint(uint64(n) + uint64(len(w.cid)))
+		writeBuf := buf.Bytes()[varint.MaxLenUvarint63-len(sizeBytes):]
+		w.buf = buf.Bytes()[varint.MaxLenUvarint63+len(w.cid):]
+		_ = copy(writeBuf[:], sizeBytes)
+
+		size := len(writeBuf)
+		if w.wo.toSkip > 0 {
```
rvagg, line 85:
```suggestion
		// indexingWriter manages state for a skip operation, but we have to mutate it here -
		// if there are still bytes to skip, then we either need to skip over this whole block, or pass
		// part of it on, and then update the toSkip state
		if w.wo.toSkip > 0 {
```

v2/internal/io/skip_writer_read_seeker.go
```diff
@@ -0,0 +1,108 @@
...
+	writeComplete chan struct{}
+}
+
+// ReWriter is a function writing to an io.Writer from an offset.
+type ReWriter func(ctx context.Context, skip uint64, w io.Writer) (uint64, error)
+
+var _ io.ReadSeeker = (*SkipWriterReaderSeeker)(nil)
+
+// NewSkipWriterReaderSeeker creates an io.ReadSeeker around a ReWriter.
+func NewSkipWriterReaderSeeker(ctx context.Context, size uint64, cons ReWriter) *SkipWriterReaderSeeker {
```
rvagg, line 30: could we have a better name than `cons`? it's not obvious to me what this name is supposed to be (`consume`?), it's puzzling when we get to `c.cons()` and it seems like it should be more obvious what it's doing - should it be read as "start consuming now"?

v2/selective.go
```diff
@@ -103,19 +128,40 @@ func TraverseToFile(ctx context.Context, ls *ipld.LinkSystem, root cid.Cid, sele
...
+	tc := traversalCar{
+		size:     conf.DataPayloadSize,
+		ctx:      ctx,
+		root:     root,
+		selector: selector,
+		ls:       ls,
+		opts:     conf,
+	}
+	rwf := func(ctx context.Context, offset uint64, writer io.Writer) (uint64, error) {
+		s, _, err := tc.WriteV1(ctx, offset, writer)
```
rvagg, line 158:
```suggestion
		// it's only at this point we have the `offset` to start writing at since the user of the
		// ReadSeeker will (may) have called Seek() and we've worked out where in the CAR
		// that is as an offset. Now we can start writing the CARv1 data, which will be passed
		// on to the ReadSeeker.
		// Note that we're inside a goroutine here
		s, _, err := tc.WriteV1(ctx, offset, writer)
```

v2/internal/io/skip_writer_read_seeker.go
```diff
@@ -0,0 +1,108 @@
...
+		writeComplete: make(chan struct{}, 1),
+	}
+}
+
+// Note: not threadsafe
+func (c *SkipWriterReaderSeeker) Read(p []byte) (n int, err error) {
+	// Check if there's already a write in progress
+	if c.reader == nil {
+		// No write in progress, start a new write from the current offset
+		// in a go routine
```
rvagg, line 44:
```suggestion
		// in a go routine and feed it back to the caller via a pipe
```

v2/selective.go
```diff
@@ -40,40 +43,62 @@ func MaxTraversalLinks(MaxTraversalLinks uint64) Option {
 	}
 }
 
+// WithDataPayloadSize sets the expected v1 size of the car being written if it is known in advance.
```
rvagg, line 46:
```suggestion
// WithDataPayloadSize sets the expected v1 size of the car being written if it is known in advance.
// This is required if NewCarV1StreamReader() is used and a Seek() operation needs seek back from
// SeekEnd (i.e. if we don't know where the end is, we can't figure out how far that is from the start).
```

v2/traversal/resumption.go
willscott: This is probably the most useful thing to have pointed out. I think the more useful method to expose here will be a `RewindToOffset` in addition to the current `RewindToPath`. I'll work on adding that

v2/selective.go
```diff
@@ -103,19 +128,40 @@ func TraverseToFile(ctx context.Context, ls *ipld.LinkSystem, root cid.Cid, sele
...
+		opts:     conf,
 	}
 
-	len, _, err := tc.WriteV1(writer)
+	len, _, err := tc.WriteV1(tc.ctx, conf.SkipOffset, writer)
 	return len, err
 }
 
+// NewCarV1StreamReader creates an io.ReadSeeker that can be used to copy out the carv1 contents of a car.
+func NewCarV1StreamReader(ctx context.Context, ls *ipld.LinkSystem, root cid.Cid, selector ipld.Node, opts ...Option) (io.ReadSeeker, error) {
```
rvagg, line 146: this should probably suggest "Selective" shouldn't it? to match `NewSelectiveWriter()`. Perhaps:

```suggestion
func NewSelectiveV1Reader(ctx context.Context, ls *ipld.LinkSystem, root cid.Cid, selector ipld.Node, opts ...Option) (io.ReadSeeker, error) {
```

rvagg:
_(FYI I messed this branch up a tiny bit so ended up rebasing it on master, removing the merge commits in the process, so the commits have changed but the resulting code is the same)_

willscott:
@rvagg are we able to merge this - was there something else we were waiting for on it?

rvagg:
Yeah, it needs #327 to get closer to working as advertised, but even then it's still not quite right. I stepped back from working on this as the complexity increased and I started questioning the utility since we mainly want this (for now) for the explore-all case and it's the selectors that make this difficult to properly do.

As it is, it can't replace Boost's impl because of caching enabled by explore-all shortcuts that we're not allowing ourselves to touch here.

I'd be fine merging in #327 to here, but I'm not sure we should merge this in since it's not going to give people what it says it does.

rvagg:
I forgot to add that I'm not intending to suggest this is a lost cause. My thinking, the last time I touched this, was that we should try to implement an efficient explore-all case here, and then iterate on that for the more complex cases.

PR #292, opened 2022-02-09
Fix install in README

Jorropo:


mvdan (approved):
Good catch - this will also work within a module.

PR #295, opened 2022-03-16
chore(deps): update dependencies of both versions

rvagg:
doing v1 and v2 here in one go

is it naughty to remove indirects? because `go get` wants to remove one of these and add 8 more if I do this without editing


mvdan (approved):
removing indirects is fine as long as `go mod tidy` is OK with it. sometimes they are redundant.

willscott:
I took the dependabot updates to codec-dagpb to latest release just now (#298 / #299  / #300 ) - is this still needed as well?

PR #296, opened 2022-03-24
Car command supports for `largebytes` nodes

willscott:
* `car ls` as an alias for `car list`
* Update to newer `unixfsnode`

### DIFF
--- a/cmd/car/get.go
+++ b/cmd/car/get.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/ipfs/go-cid"
 	ipfsbs "github.com/ipfs/go-ipfs-blockstore"
+	"github.com/ipfs/go-unixfsnode"
 	"github.com/ipld/go-car"
 	"github.com/ipld/go-car/v2/blockstore"
 	"github.com/ipld/go-ipld-prime/datamodel"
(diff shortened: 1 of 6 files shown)


cmd/car/get.go
```diff
@@ -176,7 +178,16 @@ func writeCarV2(ctx context.Context, rootCid cid.Cid, output string, bs *blockst
 		return err
 	}
 
-	err = traversalProgress.WalkMatching(rootNode, s, func(p traversal.Progress, n datamodel.Node) error { return nil })
+	err = traversalProgress.WalkMatching(rootNode, s, func(p traversal.Progress, n datamodel.Node) error {
+		lb, ok := n.(datamodel.LargeBytesNode)
+		if ok {
+			rs, err := lb.AsLargeBytes()
+			if err == nil {
+				io.Copy(io.Discard, rs)
```
mvdan, line 186: do we care about this error?

mvdan (approved):

cmd/car/get.go
willscott: the potential one from `io.Copy`, you mean?

mvdan: yup

PR #304, opened 2022-05-27
[WIP: Need to make this an option] Need to use the dagbp builder for a unixfs node

aarshkshah1992:
The 	`unixfsnode.AddUnixFSReificationToLinkSystem(ls)` dosen't work without this. A non dag pb node is rejected as a plain node by the UnixFS reirfier.

### DIFF
--- a/v2/selective.go
+++ b/v2/selective.go
@@ -12,6 +12,7 @@ import (
 	"github.com/ipld/go-car/v2/index"
 	"github.com/ipld/go-car/v2/internal/carv1"
 	"github.com/ipld/go-car/v2/internal/loader"
+	dagpb "github.com/ipld/go-codec-dagpb"
 	ipld "github.com/ipld/go-ipld-prime"
 	"github.com/ipld/go-ipld-prime/datamodel"
 	"github.com/ipld/go-ipld-prime/linking"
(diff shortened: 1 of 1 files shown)


willscott:
note that there's context in https://github.com/ipfs/go-unixfsnode/pull/27

@aschmahmann this is an example of why the IPLD side thinks that's important - when loading a node through the linksystem it will not always come through as dagPB, even when it is.

aschmahmann:
> this is an example of why the IPLD side thinks that's important - when loading a node through the linksystem it will not always come through as dagPB, even when it is.

Yeah.... I don't think the way we currently enforce strictness (i.e. insist that UnixFS data follows the spec) is great. Using type checks is just bad (e.g. what if someone made a new dagpb implementation and chose to use that with the existing ADL?). This case is still meant to follow the spec though (e.g. it's being used to transport some UnixFS data via a CAR file and it should actually be UnixFS data).

Perhaps there is more precise tooling we can use for this kind of thing though so we don't need to use concrete type checks. I could also see how for this kind of thing allowing for "loose decoding" where the producer of the CAR says "I know what I'm doing" while the consumer of the blocks from the CAR uses strict decoding to ensure the data they're getting is actually UnixFS.

willscott:
i think the less controversial immediate solution to this is in #305

BigLep:
Closing given #305

PR #305, opened 2022-05-27
PrototypeChooser support

willscott:
* being able to set an option for traversalprototypechooser seems like an oversight.
* I changed the tracked index records in the `writerOutput` to a map that would only intercept blocks on the first time it would potentially tee them to the output stream. I feel like there's something here that maybe changes based on the `BlockstoreAllowDuplicatePuts` option

cc @aarshkshah1992  - with these your example does what you want it to do.


rvagg:
The bypass for duplicate loads may have unintended consequences but the only thing I can think of is that maybe you really want to create a sloppy CAR with duplicate blocks as allowed by spec .. for some reason .. but would you be using that API if you had anything like this in mind? Maybe you should just be opting for the direct CARv1 interface instead. Aside from that, the LinkSystem is still returning the blocks if they're duplicate, just not passing down to the underlying LinkSystem. Mostly we care about not skipping duplicates when executing a complex selector and if you're doing that to watch LinkSystem loads I don't _think_ you'd care about duplicates anyway as long as you're executing the full selector.

:shrug: This might be a case of shipping it and seeing if anyone squeals.

Aside from that, the assertion on the selective test is pretty weak, is there a way to narrow down the number of expected blocks to a range? I know it's getting complicated to ask for a specific sub-dag you'd expect, as nice as that would be, but it'd also be nice to tighten up the assertion(s) a little bit to recognise more failure modes.

willscott:
added additional checks in the test case

rvagg (approved):
maybe let's get https://github.com/multiformats/go-multicodec/pull/68 sorted too so we have a tagged version here

PR #307, opened 2022-06-07
Skip offset should be a traversal option

aarshkshah1992:


v2/options.go
```diff
@@ -72,6 +73,14 @@ func ZeroLengthSectionAsEOF(enable bool) Option {
 	}
 }
 
+// WithSkipOffset sets the start offset we should seek to the in the traversal
+// when writing out a CAR.
```
masih, line 77: It would be great if the docs mention that the option only applies to selective/traversal writer

aarshkshah1992: @masih Done.

PR #309, opened 2022-06-29
Add API to regenerate index from CARv1 or CARv2

masih:
The index generation APIs either allowed reading an existing index from
a CARv2 or explicitly required a CARv1 to generate index.

Introduce APIs to make it easier for users that want to regenerate the
index regardless of whether it exists in a CAR file or not. The index
generation APIs are changed to accept either of the formats and
re-generate the index from the data payload unless `ReadOrGenerate` is
called.

Adjust the tests to run for all flavours of index generation with both
CARv1 and CARv2 payload.

### DIFF
--- a/v2/index_gen.go
+++ b/v2/index_gen.go
@@ -13,9 +13,12 @@ import (
 	"github.com/multiformats/go-varint"
 )
 
-// GenerateIndex generates index for a given car in v1 format.
+// GenerateIndex generates index for the given car payload reader.
 // The index can be stored in serialized format using index.WriteTo.
-// See LoadIndex.
+//
+// Note, the index is re-generated every time even if the payload is in CARv2 format and already has
+// an index. To read existing index when available see ReadOrGenerateIndex.
+// See: LoadIndex.
 func GenerateIndex(v1r io.Reader, opts ...Option) (index.Index, error) {
 	wopts := ApplyOptions(opts...)
 	idx, err := index.New(wopts.IndexCodec)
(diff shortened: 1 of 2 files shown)


dirkmc (approved):
LGTM 👍

v2/index_gen.go
```diff
@@ -28,21 +31,63 @@ func GenerateIndex(v1r io.Reader, opts ...Option) (index.Index, error) {
...
+			return err
+		}
 
-	// Parse Options.
-	o := ApplyOptions(opts...)
+		// Sanity-check the CARv2 header
+		if v2h.DataOffset < HeaderSize {
+			return fmt.Errorf("malformed CARv2; data offset too small: %d", v2h.DataOffset)
+		}
+		if v2h.DataSize < 1 {
```
rvagg, line 62: I think a zero-length payload is still valid isn't it? it would just produce an empty index. not a big deal, zero payload probably means an error but strictly speaking I think `0` should be acceptable.

rvagg (approved):

v2/index_gen.go
masih: I thought about this when writing this statement. I actually think `1` might be too small. The rationale is that according to CARv2 spec the data size refers to the inner CARv1; as in the entire thing including the CARv1 header. At the danger of sounding pedantic (apologies if I do 🙂) accepting zero as `DataSize` would then mean an empty file is technically a valid CARv1? 

That's why I think that value should be the minimum possible size for a CARv1, which is a CARv1 header with version value 1 and no roots.

WDYT?

masih: Looks like the CARv1 should have at [least one root](https://github.com/ipld/go-car/blob/master/car.go#L138) CID, which increases the minimum acceptable CARv1 size.

PR #312, opened 2022-07-06
Fix malformed CAR panics and excessive memory usage

masih:
Fix an issue where malformed CARv1 files cause a panic by introducing a maximum header and section size in CARv1 as an option.

Add fuzz tests across the repo


Jorropo (approved):
LGTM :crossed_fingers:

willscott:
retroactive :+1:

PR #316, opened 2022-07-06
Empty identity CID should be indexed when options are set

masih:
There is an edge-case where if the storing identity CIDs are enabled in
a CARv2, one could technically store an empty identity CID which ends up
with a `singlewidthindex` width of 8.

Such CAR files indeed exist out there and the validation changes
introduced in `2.4.0` means such CAR file indices are no longer readable
, even if regenerated.

The question is should this be considered a valid index/readable index?


masih:
Cc @nonsense @dirkmc

willscott (approved):
seems fine. not the index's problem to enforce constraints to not do silly things at other layers of the stack

PR #320, opened 2022-07-12
feat: add `car inspect` command to cmd pkg

rvagg:
I'm not sure if the `replace`s in go.mod are going to be a problem here but they seem appropriate to help building it consistently without needing to release then update here afterward


cmd/go.mod
```diff
@@ -5,70 +5,73 @@ go 1.17
...
+	golang.org/x/xerrors v0.0.0-20220609144429-65e65417b02f // indirect
+	google.golang.org/protobuf v1.28.0 // indirect
 	gopkg.in/errgo.v2 v2.1.0 // indirect
-	lukechampine.com/blake3 v1.1.6 // indirect
+	lukechampine.com/blake3 v1.1.7 // indirect
 )
+
+replace github.com/ipld/go-car/v2 => ../v2
+
+replace github.com/ipld/go-car => ../
```
masih, line 77: The rationale for not using `replace` directive is that it breaks `go install` unfortunately.

masih (approved):
My vote would be to not use `replace` directive or move it to a separate PR for further discussion. Otherwise LGTM :+1:

cmd/go.mod
willscott: yeah, would prefer to not have these replaces left in

rvagg: 👍 didn't know about `go install` .. that kind of sucks
removed the replaces

masih (approved):

Issue #314, opened 2022-07-06
Indexes are slurpy and allocates lots of memory


Jorropo:
Indexes datastructure are fully copied into memory when parsing an index.

There is no limit on how big thoses copies were (because practical indexes used by peoples are bigger than what would be a safe limit).
So if you parse an index from untrusted user input, they can send you a really big index and memory hog or OOM you.

Would be nice if we have a way to parse untrusted indexes without opening yourself to OOMs.

This was first attempted in #312 (see discussion in https://github.com/ipld/go-car-priv/pull/2):
- https://github.com/ipld/go-car/commit/3923d315f1083ce25ee4e790287cc5702a381b13

But was reverted due to API breaking considerations (see discussion in the channel and https://github.com/ipld/go-car-priv/pull/18)):
- https://github.com/ipld/go-car/commit/4bc677484af39d1153e075ae162c5ebcc6e396a3

See https://github.com/ipld/go-car/security/advisories/GHSA-9x4h-8wgm-8xfg
> *go-car@v2.4.0* also includes additional documentation regarding the dangers of consuming CARv2 index data from untrusted sources and a recommendation to regenerate indexes of CAR data from such sources where an index is required.

masih:
Once streaming index read is implemented, re-add the tests that are removed [here](https://github.com/ipld/go-car/pull/321#pullrequestreview-1036029070)

Jorropo:
@rvagg we do not use indexes in Kubo, and I have already too much things to do.
I cannot take on that work sorry.

I opened it because as far as I understood in the discussion we had indexes that allocates might be an issue, however boost was not able to do a refactor use `io.ReaderAt` instead (or / and was concerned of copying into `bytes.Buffer` due to performance issues I belive).

If consumers of indexes (*boost*) don't think this is an issue or don't care about it, then we can close this.

PR #321, opened 2022-07-12
Only read index codec during inspection

masih:
Instead of fully reading the index, only peak the index codec until
streaming index read is implemented. This would offer a safer
`Inspect()` call but with the tradeoff that overflow in index size is no
longer detected. Hence the tests that are now removed.


rvagg (approved):
it's a shame to delete those test fixtures; I note you've linked them @ https://github.com/ipld/go-car/issues/314#issuecomment-1181933896 so we're at least capturing them in the record - but since we're saying there that we're wanting to bring them back, they could just be commented out for now, but I'll leave that up to you

PR #323, opened 2022-07-27
blockstore: allow to pass a file to write in

MichaelMure:
This allow the caller to control the file lifecycle and implement a cleaning strategy. An example would be:
- open a file in a temporary folder
- if the write suceed, close the file and move it to its final destination
- on error, remove the file

If the underlying filesystem support it, an anonymous file can be used instead: open then immediately delete, use the file descriptor to write. A crash before the end of the write process would release the used storage automatically.

Not a breaking change, only introduce an extra function.

### DIFF
--- a/v2/blockstore/readwrite.go
+++ b/v2/blockstore/readwrite.go
@@ -10,12 +10,13 @@ import (
 	blocks "github.com/ipfs/go-block-format"
 	"github.com/ipfs/go-cid"
 	blockstore "github.com/ipfs/go-ipfs-blockstore"
+	"github.com/multiformats/go-varint"
+
 	carv2 "github.com/ipld/go-car/v2"
 	"github.com/ipld/go-car/v2/index"
 	"github.com/ipld/go-car/v2/internal/carv1"
 	"github.com/ipld/go-car/v2/internal/carv1/util"
 	internalio "github.com/ipld/go-car/v2/internal/io"
-	"github.com/multiformats/go-varint"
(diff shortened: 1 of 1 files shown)


Jorropo (approved):
Imo we should rename `OpenReadWrite` to `OpenReadWritePath` and `OpenReadWriteFile` to `OpenReadWrite` but I don't think it's worth uselessly breaking consumers.

LGTM

v2/blockstore/readwrite.go
```diff
@@ -103,6 +104,18 @@ func OpenReadWrite(path string, roots []cid.Cid, opts ...carv2.Option) (*ReadWri
...
+	if err != nil {
+		return nil, err
+	}
+	// close the file when finalizing
+	rwbs.ronly.carv2Closer = rwbs.f
+	return rwbs, nil
+}
+
+// OpenReadWriteFile is similar as OpenReadWrite but let you control the file lifecycle.
+// You are responsible to close the given file.
```
rvagg, line 117:
```suggestion
// OpenReadWriteFile is similar as OpenReadWrite but lets you control the file lifecycle.
// You are responsible for closing the given file.
```

rvagg (approved):
👍

rvagg:
aaaaactually, I probably should have asked for tests - @MichaelMure would you mind taking a look at the current test suite and see if you can extend one of them to control the file from outside rather than passing the path?

PR #324, opened 2022-08-01
OpenReadWriteFile: add test

MichaelMure:
Address https://github.com/ipld/go-car/pull/323#issuecomment-1200823708

### DIFF
--- a/v2/blockstore/readwrite_test.go
+++ b/v2/blockstore/readwrite_test.go
@@ -945,6 +945,41 @@ func TestReadWrite_ReWritingCARv1WithIdentityCidIsIdenticalToOriginalWithOptions
 	require.Equal(t, wantSum, gotSum)
 }
 
+func TestReadWriteOpenFile(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+
+	dir := t.TempDir() // auto cleanup
+	f, err := ioutil.TempFile(dir, "")
+	require.NoError(t, err)
+
+	root := blocks.NewBlock([]byte("foo"))
+
+	bs, err := blockstore.OpenReadWriteFile(f, []cid.Cid{root.Cid()})
+	require.NoError(t, err)
+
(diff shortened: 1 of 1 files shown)


v2/blockstore/readwrite_test.go
```diff
@@ -945,6 +945,38 @@ func TestReadWrite_ReWritingCARv1WithIdentityCidIsIdenticalToOriginalWithOptions
...
+	_, err = bs.Get(ctx, roots[0])
+	require.NoError(t, err)
+	_, err = bs.GetSize(ctx, roots[0])
+	require.NoError(t, err)
+
+	err = bs.Finalize()
+	require.NoError(t, err)
+
+	_, err = f.Seek(0, 0)
+	require.NoError(t, err) // file should not be closed, let the caller do it
```
masih, line 977: The test is the caller, and therefore should close the file. No?

MichaelMure: at least windows seems to agree with you, I'll add that

v2/index/index.go
```diff
@@ -102,6 +102,13 @@ func GetFirst(idx Index, key cid.Cid) (uint64, error) {
 	return firstOffset, err
 }
 
+// func Has(idx Index, key cid.Cid) (bool, error) {
+// 	has := false
+// 	err := idx.GetAll(key, func(_ uint64) bool {
+//
+// 	})
+// }
+
```
masih, line 111: It seems the scope of this PR expanded since my last comment. Is that intended?

masih (changes requested):

v2/index/index.go
MichaelMure: my fingers decided to type `git commit -a` after I carefully selected my changes ...

masih (approved):

v2/index/index.go
masih: no worries thank you

MichaelMure:
FYI I don't have the power to merge.

PR #329, opened 2022-08-23
feat: index identity multihash roots as 0

rvagg:
Exploring options for https://github.com/filecoin-project/boost/pull/715, this is just some code to test out an option, no need to take this seriously yet.

### DIFF
--- a/v2/index_gen.go
+++ b/v2/index_gen.go
@@ -41,13 +41,13 @@ func LoadIndex(idx index.Index, r io.Reader, opts ...Option) error {
 	o := ApplyOptions(opts...)
 
 	reader := internalio.ToByteReadSeeker(r)
-	pragma, err := carv1.ReadHeader(r, o.MaxAllowedHeaderSize)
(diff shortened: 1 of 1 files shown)


v2/index_gen.go
```diff
@@ -156,6 +173,12 @@ func LoadIndex(idx index.Index, r io.Reader, opts ...Option) error {
 		}
 	}
 
+	for r, store := range identityRoots {
+		if store {
+			records = append(records, index.Record{Cid: r, Offset: uint64(1)})
```
rvagg, line 178: I'd rather this be `0` but it can't be because of https://github.com/filecoin-project/index-provider/blob/b918da7508b05263988561ec3f28ca31e0fcacb9/multihash_iterator.go#L48-L52

But also, is it really something we should report to the indexer, perhaps it should get filtered out before it gets there. :shrug:

rvagg:
not proceeding with this approach

PR #332, opened 2022-09-02
feat: Has() and Get() will respect StoreIdentityCIDs option

rvagg:
When StoreIdentityCIDs is set, it will defer to the index to check whether the
blocks are in the CAR. When the CAR is a v1 and StoreIdentityCIDs is set, the
index will contain the identity CIDs. When it's a v2 with an existing index,
however that index was created will determine whether the identity CIDs are
that are in the CAR are found.

When StoreIdentityCIDs is not set, Has() will always return true and Get() will
always return the block.

---

The background to this is trying to get identity CID retrievals properly working in Lotus _and_ trying to emulate the `car.SelectiveCar` CAR creation when using the RW blockstore to back a Graphsync transfer. Currently, a Lotus client will use `car.SelectiveCar` to make the CAR for the deal to calculate CommP. Then it'll transfer the CAR to the server, but via Graphsync, the server will not include identity CIDs, so the CommP will not match.

The primary purpose of respecting `StoreIdentityCIDs` for the blockstore interface is so that when backing Graphsync, `Has()` will report correctly whether the block has been stored or not, so that `Put()` will properly write the block into the CAR. Note that `Put()` (actually `PutMany()`) will already respect `StoreIdentityCIDs`, so no change needed there; but `Has()` needs to not return `true` for the block to be put in the first place.

There's two alternatives I think worth considering:

1. Introducing a new option here to handle this functionality, something not called "Store". A new name could help align the slight confusion that may arise in the existing-index case.
2. Entirely ditching `car.SelectiveCar` as the primary method of making CARs for online deals in Lotus—unfortunately this method is widespread, and there's even code in the wild that emulates it! (e.g. https://github.com/filecoin-project/lotus/blob/a843c52e38da13da489cbe6b290ea49b2660b3fb/node/impl/client/client.go#L995). My inclination is to say that this ship sailed when Filecoin was launched - online deals require that identity CIDs appear in the CAR on both sides for CommP to match. This just needs to be fixed on the server side since it was broken when the CARv2 blockstore interface was pushed through Lotus, and we should also bake it into the Filecoin spec perhaps.

### DIFF
--- a/v2/blockstore/readonly.go
+++ b/v2/blockstore/readonly.go
@@ -217,14 +217,23 @@ func (b *ReadOnly) DeleteBlock(_ context.Context, _ cid.Cid) error {
 }
 
 // Has indicates if the store contains a block that corresponds to the given key.
-// This function always returns true for any given key with multihash.IDENTITY code.
+// This function always returns true for any given key with multihash.IDENTITY
+// code unless the StoreIdentityCIDs option is on, in which case it will defer
+// to the index to check for the existence of the block; the index may or may
+// not contain identity CIDs included in this CAR, depending on whether
+// StoreIdentityCIDs was on when the index was created. If the CAR is a CARv1
+// and StoreIdentityCIDs is on, then the index will contain identity CIDs and
+// this will always return true.
 func (b *ReadOnly) Has(ctx context.Context, key cid.Cid) (bool, error) {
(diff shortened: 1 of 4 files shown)


rvagg:
I'll merge and tag this as v2.5.0 tomorrow unless there are objections

masih (approved):
Apologies for the delay in reviewing this

PR #334, opened 2022-09-22
Add read buffers wherever ReadByte() is called

dirkmc:


willscott:
would there be a way to avoid this extra layer of bufio copies for cases where the backing ReaderAt passed to the car library is already a memory buffer?

PR #337, opened 2022-10-18
Expose iterator over a CAR

dirkmc:
The iterator emits the section cid, section offset and block data length for each section in the CAR


dirkmc:
Superseded by https://github.com/ipld/go-car/pull/338

PR #338, opened 2022-10-19
add a `SkipNext` method on block reader

willscott:
alternative to #337

* [x] Confirm interface meets piece store index needs.
* [x] Tests

### DIFF
--- a/v2/block_reader.go
+++ b/v2/block_reader.go
@@ -9,6 +9,7 @@ import (
 	"github.com/ipld/go-car/v2/internal/carv1"
 	"github.com/ipld/go-car/v2/internal/carv1/util"
 	internalio "github.com/ipld/go-car/v2/internal/io"
+	"github.com/multiformats/go-varint"
 )
 
 // BlockReader facilitates iteration over CAR blocks for both CARv1 and CARv2.
@@ -20,8 +21,9 @@ type BlockReader struct {
 	Roots []cid.Cid
 
 	// Used internally only, by BlockReader.Next during iteration over blocks.
-	r    io.Reader
-	opts Options
+	r      io.Reader
+	offset uint64
+	opts   Options
 }
 
 // NewBlockReader instantiates a new BlockReader facilitating iteration over blocks in CARv1 or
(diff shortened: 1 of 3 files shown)


v2/block_reader.go
```diff
@@ -122,5 +129,88 @@ func (br *BlockReader) Next() (blocks.Block, error) {
...
+// If the underlying reader used by the BlockReader is actually a ReadSeeker, this method will attempt to
+// seek over the underlying data rather than reading it into memory.
+func (br *BlockReader) SkipNext() (*BlockMetadata, error) {
+	sctSize, err := util.LdReadSize(br.r, br.opts.ZeroLengthSectionAsEOF, br.opts.MaxAllowedSectionSize)
+	if err != nil {
+		return nil, err
+	}
+
+	if sctSize == 0 {
+		_, _, err := cid.CidFromBytes([]byte{})
```
rvagg, line 155: Props for this hack to get the right error msg

v2/block_reader.go
```diff
@@ -122,5 +129,88 @@ func (br *BlockReader) Next() (blocks.Block, error) {
...
+
+	cidSize, c, err := cid.CidFromReader(io.LimitReader(br.r, int64(sctSize)))
+	if err != nil {
+		return nil, err
+	}
+
+	blkSize := sctSize - uint64(cidSize)
+
+	if brs, ok := br.r.(io.ReadSeeker); ok {
+		if br.readerSize == -1 {
```
rvagg, line 167:
```suggestion
		// carv1 and we don't know the size, so work it out and cache it
		if br.readerSize == -1 {
```

v2/block_reader.go
```diff
@@ -122,5 +129,88 @@ func (br *BlockReader) Next() (blocks.Block, error) {
...
+		br.offset = uint64(finalOffset)
+		return &BlockMetadata{
+			c,
+			uint64(finalOffset) - sctSize - uint64(varint.UvarintSize(sctSize)),
+			blkSize,
+		}, nil
+	}
+
+	// read to end.
+	_, err = io.CopyN(io.Discard, br.r, int64(blkSize))
```
rvagg, line 201: Any reason to not check the `written` value against expected, like we're doing for the seek version above?

rvagg (approved):
lgtm if it serves purpose, although I'm a little skeptical that the skipping is going to save much aside from passing slices around given the likely layering of buffered reads down to the OS and the size of our blocks being relatively small. But maybe that's worth it?

v2/block_reader.go
willscott: copy-to-discard should be able to not allocate a buffer when moving the reader forward

rvagg: but it still returns a number of bytes "copied", right? so checking that the return value matches `blkSize` seems like an appropriate thing to do?

PR #341, opened 2022-11-07
Add a debugging form for car files.

willscott:
This change adds two new sub-commands to the car CLI

car debug file.car

creates a patch-file-compatible representation of the content of the car file. Blocks will be represented in dag-json pretty-printed form.

car compile file.patch

will do the inverse process of building a car file from a debug patch file. CIDs will be re-compiled based on the contents of blocks, with links in parent blocks updated to point to the compiled values.

an example debug patch of the car used in the testscript test fixture would be:

```patch
car compile --v2 small.car
root bafybeidx5vxxny6ca3mgs5d6wy5ubwcibpirctmktpkvfk4io34i2ww2hy
for raw: bytes are 68656c6c6f20776f726c640a
--- bafkreifjjcie6lypi6ny7amxnfftagclbuxndqonfipmb64f2km2devei4
+++ raw bafkreifjjcie6lypi6ny7amxnfftagclbuxndqonfipmb64f2km2devei4
@@ -0,1 +0,1 @@
hello world

--- bafybeidx5vxxny6ca3mgs5d6wy5ubwcibpirctmktpkvfk4io34i2ww2hy
+++ json (no-end-cr) bafybeidx5vxxny6ca3mgs5d6wy5ubwcibpirctmktpkvfk4io34i2ww2hy
@@ -0,15 +0,15 @@
{
  "Data": {
    "/": {
      "bytes": "CAE"
    }
  },
  "Links": [
    {
      "Hash": {
        "/": "bafkreifjjcie6lypi6ny7amxnfftagclbuxndqonfipmb64f2km2devei4"
      },
      "Name": "foo.txt",
      "Tsize": 12
    }
  ]
}
```

### DIFF
--- a/cmd/car/compile.go
+++ b/cmd/car/compile.go
@@ -4,7 +4,6 @@ import (
 	"bufio"
 	"bytes"
 	"context"
-	"encoding/json"
 	"fmt"
 	"io"
 	"os"
@@ -19,13 +18,19 @@ import (
 	carv2 "github.com/ipld/go-car/v2"
 	"github.com/ipld/go-car/v2/blockstore"
 	"github.com/ipld/go-ipld-prime"
+	"github.com/ipld/go-ipld-prime/codec"
 	"github.com/ipld/go-ipld-prime/codec/dagjson"
 	"github.com/ipld/go-ipld-prime/datamodel"
 	"github.com/ipld/go-ipld-prime/linking"
 	cidlink "github.com/ipld/go-ipld-prime/linking/cid"
 	"github.com/ipld/go-ipld-prime/node/basicnode"
 	"github.com/ipld/go-ipld-prime/storage/memstore"
+	"github.com/polydawn/refmt/json"
 	"github.com/urfave/cli/v2"
+)
+
+var (
+	plusLineRegex = regexp.MustCompile(`^\+\+\+ ([\w-]+) ([\S]+ )?([\w]+)$`)
 )
 
 // Compile is a command to translate between a human-debuggable patch-like format and a car file.
@@ -46,10 +51,12 @@ func CompileCar(c *cli.Context) error {
 		return err
 	}
 
-	v2 := strings.Contains(string(header), "--v2")
-	trimH := strings.TrimSpace(string(header))
-	headerParts := strings.Split(trimH, " ")
-	carName := headerParts[len(headerParts)-1]
+	v2 := strings.HasPrefix(string(header), "car compile --v2 ")
+	rest := strings.TrimPrefix(string(header), "car compile ")
+	if v2 {
+		rest = strings.TrimPrefix(rest, "--v2 ")
+	}
+	carName := strings.TrimSpace(rest)
 
 	roots := make([]cid.Cid, 0)
 	for {
@@ -209,7 +216,7 @@ func serializeBlock(ctx context.Context, codec cid.Prefix, encoding string, raw 
 	ls.SetReadStorage(&store)
 	ls.SetWriteStorage(&store)
 	b := basicnode.Prototype.Any.NewBuilder()
-	if encoding == "json" {
+	if encoding == "dag-json" {
 		if err := dagjson.Decode(b, bytes.NewBuffer(raw)); err != nil {
 			return cid.Undef, nil, err
 		}
@@ -301,20 +308,19 @@ func patch(ctx context.Context, c cid.Cid, blk []byte) ([]byte, error) {
 		return nil, fmt.Errorf("could not load block: %q", err)
 	}
 
-	outMode := "json"
+	outMode := "dag-json"
 	if node.Kind() == datamodel.Kind_Bytes && isPrintable(node) {
 		outMode = "raw"
 	}
 	finalBuf := bytes.NewBuffer(nil)
 
-	if outMode == "json" {
-
-		initalJson := bytes.NewBuffer(nil)
-		if err := dagjson.Encode(node, initalJson); err != nil {
-			return nil, err
+	if outMode == "dag-json" {
+		opts := dagjson.EncodeOptions{
+			EncodeLinks: true,
+			EncodeBytes: true,
+			MapSortMode: codec.MapSortMode_Lexical,
 		}
-		// re-do it with standard json to pretty print it.
-		if err := json.Indent(finalBuf, initalJson.Bytes(), "", "  "); err != nil {
+		if err := dagjson.Marshal(node, json.NewEncoder(finalBuf, json.EncodeOptions{Line: []byte{'\n'}, Indent: []byte{'\t'}}), opts); err != nil {
 			return nil, err
 		}
 	} else if outMode == "raw" {
@@ -352,6 +358,10 @@ func isPrintable(n ipld.Node) bool {
 	if bytes.ContainsAny(b, string([]byte{0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x10, 0x11, 0x12, 0x13, 0x14, 0x16, 0x17, 0x18, 0x19, 0x1c, 0x1d, 0x1e, 0x1f})) {
 		return false
 	}
+	// check if would confuse the 'end of patch' checker.
+	if bytes.Contains(b, []byte("\n--- ")) {
+		return false
+	}
 	return true
 }
 
@@ -377,13 +387,12 @@ func parsePatch(br *bufio.Reader) (cid.Cid, string, []byte, error) {
 	}
 	var mode string
 	var noEndReturn bool
-	r := regexp.MustCompile(`^\+\+\+ ([\w]+) ([\S]+ )?([\w]+)$`)
-	matches := r.FindSubmatch(l2)
+	matches := plusLineRegex.FindSubmatch(l2)
 	if len(matches) >= 2 {
 		mode = string(matches[1])
 	}
 	if len(matches) < 2 || string(matches[len(matches)-1]) != cs {
-		return cid.Undef, "", nil, fmt.Errorf("mismatched cid lines")
+		return cid.Undef, "", nil, fmt.Errorf("mismatched cid lines: %v", string(l2))
 	}
 	if len(matches[2]) > 0 {
 		noEndReturn = (string(matches[2]) == "(no-end-cr) ")
@@ -427,9 +436,9 @@ func parsePatch(br *bufio.Reader) (cid.Cid, string, []byte, error) {
 		}
 	}
 
-	// remove the final line return
 	ob := outBuf.Bytes()
 
+	// remove the final line return
 	if len(ob) > 2 && bytes.Equal(ob[len(ob)-2:], []byte("\r\n")) {
 		ob = ob[:len(ob)-2]
 	} else if len(ob) > 1 && bytes.Equal(ob[len(ob)-1:], []byte("\n")) {
--- a/cmd/car/compile.go
+++ b/cmd/car/compile.go
@@ -92,7 +92,7 @@ func CompileCar(c *cli.Context) error {
 		rawCodecs[nextCid] = mode
 	}
 
-	fmt.Printf("structuring as tree...\n")
+	//fmt.Printf("structuring as tree...\n")
 	// structure as a tree
 	childMap := make(map[cid.Cid][]cid.Cid)
 	for c := range rawBlocks {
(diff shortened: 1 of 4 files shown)


b5:
Fully support having this!

rvagg:
windows failures, probably crlf related?

also this is weird on macos https://github.com/ipld/go-car/actions/runs/3410523283/jobs/5673582480, maybe just a flake since it's passing on the other 3 macos runners but it seems cmd related

willscott:
There's some sort of flakiness with testscript in general.

@rvagg  are you okay with design / the proposed debug patch format modulo the test?

cmd/car/compile.go
```diff
@@ -0,0 +1,446 @@
...
+		if err != nil {
+			return err
+		}
+		defer outFile.Close()
+		outStream = outFile
+	}
+
+	outStream.WriteString("car compile ")
+	if rd.Version == 2 {
+		outStream.WriteString("--v2 ")
```
rvagg, line 264: It looks like this is the only hint that it once was a v2 and there's not space to say anything else about it? I guess mostly we care about stability of v1 forms and the v2 is just a convenience wrapper, but if we went ahead with additional features, such as the messaging capability in https://github.com/ipld/go-car/pull/322, then where would we put these things? Would it be hard to extend this format to include those things, and could we do it in a non-breaking way perhaps?

cmd/car/compile.go
```diff
@@ -0,0 +1,446 @@
...
+	if outMode == "json" {
+
+		initalJson := bytes.NewBuffer(nil)
+		if err := dagjson.Encode(node, initalJson); err != nil {
+			return nil, err
+		}
+		// re-do it with standard json to pretty print it.
+		if err := json.Indent(finalBuf, initalJson.Bytes(), "", "  "); err != nil {
+			return nil, err
+		}
```
rvagg, line 319: so .. this is a bit naughty but you can pull this off in one pass with go-ipld-prime by calling `Marshal` and passing in some refmt options, see https://go.dev/play/p/-OCdX_w5l52

cmd/car/compile.go
```diff
@@ -0,0 +1,446 @@
...
+	}
+
+	//parse headers.
+	br := bufio.NewReader(inStream)
+	header, _, err := br.ReadLine()
+	if err != nil {
+		return err
+	}
+
+	v2 := strings.Contains(string(header), "--v2")
```
rvagg, line 49: probably need to do more than this because the file name could contain this

```suggestion
	v2 := strings.Index(string(header), "car compile --v2 ") != -1
```
?

cmd/car/compile.go
```diff
@@ -0,0 +1,446 @@
...
+	trimH := strings.TrimSpace(string(header))
+	headerParts := strings.Split(trimH, " ")
+	carName := headerParts[len(headerParts)-1]
```
rvagg, line 52: doesn't allow for spaces in filenames, which is a thing, perhaps split off `car compile ` then optionally `--v2 `.. Future extensions are going to have to deal with `--foobar` and argument ordering but for now it can be simple

cmd/car/compile.go
```diff
@@ -0,0 +1,446 @@
...
+			return err
+		}
+		if bytes.Equal(peek, []byte("--- ")) {
+			break
+		}
+		rootLine, _, err := br.ReadLine()
+		if err != nil {
+			return err
+		}
+		if strings.HasPrefix(string(rootLine), "root ") {
```
rvagg, line 69: ok, so this is is not exhaustive, if it doesn't match `root ` then it continues to loop and drop `rootLine`. So is this where we get potential backward compatibility of additional v2 features that we can insert with `---`, and also the `--foobar` arguments in the header?

rvagg: (continuing my thought below about v2 features, where I started my comments on debug first before moving up here to compile)

cmd/car/compile.go
```diff
@@ -0,0 +1,446 @@
...
+	l2, isPrefix, err := br.ReadLine()
+	if err != nil {
+		return cid.Undef, "", nil, err
+	}
+	if isPrefix {
+		return cid.Undef, "", nil, fmt.Errorf("unexpected long header l2")
+	}
+	var mode string
+	var noEndReturn bool
+	r := regexp.MustCompile(`^\+\+\+ ([\w]+) ([\S]+ )?([\w]+)$`)
```
rvagg, line 380: perhaps pull this up to the top of the file

cmd/car/compile.go
```diff
@@ -0,0 +1,446 @@
...
+	store := memstore.Store{Bag: map[string][]byte{}}
+	ls.SetReadStorage(&store)
+	ls.SetWriteStorage(&store)
+	store.Put(ctx, c.KeyString(), blk)
+	node, err := ls.Load(linking.LinkContext{Ctx: ctx}, cidlink.Link{Cid: c}, basicnode.Prototype.Any)
+	if err != nil {
+		return nil, fmt.Errorf("could not load block: %q", err)
+	}
+
+	outMode := "json"
```
rvagg, line 304: maybe we should make this `"dag-json"` to be clear about what's going on and give optionality in the future--if we're specific about codec now then we have more flexibility later

while it's true that this is not strict dag-json because it's pretty-printed, it's still dag-json compatible for decoders.

cmd/car/compile.go
```diff
@@ -0,0 +1,446 @@
...
+		if err == io.EOF {
+			break
+		} else if err != nil {
+			return cid.Undef, "", nil, err
+		}
+	}
+
+	// remove the final line return
+	ob := outBuf.Bytes()
+
```
rvagg, line 432:
```suggestion
	ob := outBuf.Bytes()

	// remove the final newline
```

cmd/car/compile.go
```diff
@@ -0,0 +1,446 @@
...
+		} else if err != nil {
+			return err
+		}
+		rawBlocks[nextCid] = nextBlk
+		rawCodecs[nextCid] = mode
+	}
+
+	//fmt.Printf("structuring as tree...\n")
+	// structure as a tree
+	childMap := make(map[cid.Cid][]cid.Cid)
```
rvagg, line 97: whoa, so I think this whole next section (x2) exists because we're not confident that the resulting CIDs will match the input CIDs, so we're doing a search, reconstruct, replace operation on them all, is that right?

  1. why do we not have confidence they're going to reconstruct byte-perfect, shouldn't that be reasonable? There might be a CIDv0 CIDv1 difference but we could easily do that check -- if expected CID is v0 then downcast actual CID and compare.
  2. why don't we just error if the resulting CID of the reconstructed block doesn't match the expected?

cmd/car/compile.go
rvagg:
```suggestion
	v2 := strings.HasPrefix(string(header), "car compile --v2 ")
```
that's probably better eh?

rvagg (changes requested):
This is probably OK, I'd like to know what itch was being scratched here? it's certainly a nice utility to peek inside a CAR and perhaps even a nice way to peek inside IPLD blocks in general without having to code something up (you could even `ipfs export .. | car debug -` to get pretty-printed output that you can't get from `ipfs dag get --output-codec=dagjson`, but that's quite a hack!).

Aside from comments inline, the main concern I have is the ability to corrupt the format if you include a `raw` block that is diff-like. You could even end up doing that with this tool - make a .patch from a CAR, bundle that up as unixfs and include it in a CAR and make a .patch from that and :boom: it will break because `--- `. But I think you could easily address that by changing `isPrintable()` to also check for `^--- ` and `false` out if there's one in there.

willscott:
> This is probably OK, I'd like to know what itch was being scratched here? it's certainly a nice utility to peek inside a CAR and perhaps even a nice way to peek inside IPLD blocks in general without having to code something up (you could even `ipfs export .. | car debug -` to get pretty-printed output that you can't get from `ipfs dag get --output-codec=dagjson`, but that's quite a hack!).
> 

for example:
https://twitter.com/pfrazee/status/1589747431071428609
https://gitlab.com/bnewbold/adenosine/-/blob/main/notes/ipld_car_explore.md

cmd/car/compile.go
willscott: this is mirroring [`git patch`](https://www.oreilly.com/library/view/git-pocket-guide/9781449327507/ch11.html)` where the first line mirrors the command used to generate the patch.

Here we use v2 to indicate if the original car was a v2 or not, and can use that as a default to re-build the same car format if it is not specified explicitly on the command line when re-compiling

cmd/car/compile.go
rvagg: actually it's slightly more complicated to make proper dag-json, I forgot the options to encode links, bytes and sort maps: https://go.dev/play/p/XyfcQAzhzpw

cmd/car/compile.go
rvagg: Thought about this after writing and I realise that it's probably because we anticipate some input to come in badly encoded forms so round-tripping is going to result in mismatched CIDs. So I guess that's why this is here.

This does seem like a lot of effort to go to though; and it's also a little error-prone just replacing the CIDs as strings. That won't necessarily always find the actual links, they could just be included as text and it assumes that they want them to be changed. A find and replace should probably be at least looking for `"/": "...."\n`, but even that's not necessarily accurate either. A more complete approach would be to walk the instantiated data model form and change the links out, but I guess that's even more code complication! Then there's the CIDv0 vs CIDv1 thing, what if the original wants to be in CIDv0 but we up-convert them to CIDv1?

Lots of effort to do it the _right_ way, but it makes me question whether doing it at all is worth it.

Your call I guess, but maybe add some comments in here about what it's doing so the first person to encounter a bug with this knows why and can choose whether to fix it. It wasn't clear to me what it was doing until I walked through the whole two blocks.

willscott: this is for the 'i want to mutate the graph and re-build a valid tree' case.
if i change the json in the patch, then that block will have a new cid when it's hashed.
that hash change needs to propagate back up the DAG.

data model still won't get all edge cases - what if it's a different codec for the same MH? what if the link is encoded in a block that's raw or that we can't parse?

this is the first pass that works reasonably well for the json/cbor cases that i've attempted to manually edit.

i am unconvinced it's worth the time to do much more complex work and still not do something perfect vs where it is currently, which supports a pretty valuable use case already.

cmd/car/compile.go
willscott: yes, other meta info would go as lines in this section.

willscott:
made the raw blocks a bit more cautious per the edge case you pointed out

cmd/car/compile.go
rvagg: oh right, tinker with the content and regenerate the graph; fair enough, I can see that being useful - it does seem like something that needs clear caveats in comments though!

cmd/car/compile.go
```diff
@@ -0,0 +1,455 @@
...
+			outStream = outFile
+		}
+
+		if err := carv1.WriteHeader(&carv1.CarHeader{
+			Roots:   roots,
+			Version: 1,
+		}, outStream); err != nil {
+			return err
+		}
+		for c, blk := range outBlocks {
```
rvagg, line 185: the failing tests are because of this, and the same block below for v1 -- the blocks are ending up shuffled thanks to iterating over the go `map[]` .. we're going to need to keep a slice of CIDs to iterate over and then rewrite those when you do the reconstruction.

rvagg (changes requested):

willscott:
good catch, @rvagg - made the output match the order of initial blocks.

cmd/car/compile.go
```diff
@@ -0,0 +1,461 @@
...
+			} else if strings.Contains(string(blk), string(c.Bytes())) {
+				if _, ok := childMap[d]; !ok {
+					childMap[d] = make([]cid.Cid, 0)
+				}
+				childMap[d] = append(childMap[d], c)
+			}
+		}
+	}
+
+	//fmt.Printf("rebuilding...\n")
```
rvagg, line 130:
```suggestion
```

cmd/car/compile.go
```diff
@@ -0,0 +1,461 @@
...
+			break
+		} else if err != nil {
+			return err
+		}
+		rawBlocks[nextCid] = nextBlk
+		rawCodecs[nextCid] = mode
+		cidList = append(cidList, nextCid)
+	}
+
+	//fmt.Printf("structuring as tree...\n")
```
rvagg, line 105:
```suggestion
	// Re-create the original IPLD encoded blocks, but allowing for modifications of the
	// patch data which may generate new CIDs; so we track the DAG relationships and
	// rewrite CIDs in other referring where they get updated.

```

rvagg (commented):
nice green ticks

I still think it needs notes in the DAG rewrite bit, so I've added my suggestions for what that might look like.

rvagg (approved):

PR #350, opened 2023-01-16
feat(cmd): add index create subcommand to create an external carv2 index

aschmahmann:
Ran into a case where I had a large CARv1 and wanted to create an index from it so I could use the two as a CARv2 blockstore (trying to take a 120GB ~69M block CARv1 of Filecoin state and use it as a blockstore for queries). IIUC this separation of the CARv1 from the index is common in Lotus/Boost and is part of why there's already a `detach` command.

First time contributing to this CLI so lmk if there's a better way you had in mind, or a better place to put this command. We may also want an `index attach` command at some point, but that can be a separate PR.

If people are happy with this, I can add a basic test for this as well (once I figure out the syntax of the test framework).

### DIFF
--- a/cmd/car/testdata/script/index-create.txt
+++ b/cmd/car/testdata/script/index-create.txt
@@ -1,0 +1,3 @@
+car index create ${INPUTS}/sample-v1.car sample-v1.car.idx
+car detach-index ${INPUTS}/sample-wrapped-v2.car sample-wrapped-v2.car.idx
+cmp sample-v1.car.idx sample-wrapped-v2.car.idx
--- a/cmd/car/car.go
+++ b/cmd/car/car.go
@@ -153,6 +153,11 @@ func main1() int {
 						Usage: "Write output as a v1 or v2 format car",
 					},
 				},
+				Subcommands: []*cli.Command{{
+					Name:   "create",
+					Usage:  "Write out a detached index",
+					Action: CreateIndex,
+				}},
 			},
 			{
 				Name:   "inspect",
--- a/cmd/car/index.go
+++ b/cmd/car/index.go
@@ -166,3 +166,45 @@ func IndexCar(c *cli.Context) error {
 	_, err = index.WriteTo(idx, outStream)
 	return err
 }
+
+// CreateIndex is a command to write out an index of the CAR file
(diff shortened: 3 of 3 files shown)


willscott:
I'm in favor of this.

the CLI commands are really just a grab bag of things we've found useful through development. at some point we should probably go through and clean it up with an eye towards what users would actually want now.

That said, this seems like the right place to add this functionality. If you add a basic test I'm happy 👍

rvagg:
👌 simple test for this shouldn't be too hard

cmd/car/testdata/script/index-create.txt
```diff
@@ -0,0 +1,3 @@
+car index create ${INPUTS}/sample-v1.car sample-v1.car.idx
+car detach-index ${INPUTS}/sample-wrapped-v2.car sample-wrapped-v2.car.idx
+cmp sample-v1.car.idx sample-wrapped-v2.car.idx
```
aschmahmann, line 3: lmk if this is too simple and I can either add an index to the inputs to compare against or add something like:

```
car detach-index list sample-v1.car.idx
cmp stdout index-list.txt

-- index-list.txt --
...
```

rvagg: no, I think this is fine; we're not exactly doing sophisticated and high-coverage tests here in cmd

this does make me think though that maybe `detach` should be moved to a subcmd of `index` too

rvagg (approved):

rvagg:
macos on 0.18 has been flaky for a little while now, unrelated to this change. gtg

cmd/car/testdata/script/index-create.txt
aschmahmann: > this does make me think though that maybe detach should be moved to a subcmd of index too

I agree, but figured I'd let the maintainers make that call and if they'd be ok with a breaking move or want both supported.

rvagg: we don't exactly have an SLA on this, cmd is currently just a grab-bag of utils we collectively find useful

aschmahmann: SGTM. Happy to submit a PR to move it once this one is closed out

Issue #354, opened 2023-01-19
Create executable binaries for every release


davidd8:
Add a github workflow to generate executable binaries for every major platform, much like https://github.com/filecoin-project/lassie/blob/main/.github/workflows/release-binaries.yml in order to make it easier for end users to run go-car out of the box.

PR #355, opened 2023-01-25
feat: add release-binaries workflow

rvagg:
Closes: https://github.com/ipld/go-car/issues/354

### DIFF
--- a/.github/workflows/release-binaries.yml
+++ b/.github/workflows/release-binaries.yml
@@ -1,29 +1,26 @@
+name: GoReleaser
 on:
   release:
-    types: [created]
+    types: [ published ]
 
 jobs:
-  releases-matrix:
-    name: Release Go Binary
+  bin-releaser:
+    name: Release Binaries
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        # build and publish in parallel: linux/386, linux/amd64, linux/arm64, windows/386, windows/amd64, darwin/amd64, darwin/arm64
-        goos: [linux, windows, darwin]
-        goarch: ["386", amd64, arm64]
-        exclude:
-          - goarch: "386"
-            goos: darwin
-          - goarch: arm64
-            goos: windows
     steps:
-    - uses: actions/checkout@v3
-    - uses: wangyoucao577/go-release-action@v1.34
-      with:
-        github_token: ${{ secrets.GITHUB_TOKEN }}
-        goos: ${{ matrix.goos }}
-        goarch: ${{ matrix.goarch }}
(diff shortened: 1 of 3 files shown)


.github/workflows/release-binaries.yml
```diff
@@ -0,0 +1,29 @@
...
+    steps:
+    - uses: actions/checkout@v3
+    - uses: wangyoucao577/go-release-action@v1.34
+      with:
+        github_token: ${{ secrets.GITHUB_TOKEN }}
+        goos: ${{ matrix.goos }}
+        goarch: ${{ matrix.goarch }}
+        goversion: "https://go.dev/dl/go1.19.5.linux-amd64.tar.gz"
+        project_path: "./cmd/lassie"
+        binary_name: "lassie"
```
willscott, line 28: these change?

masih (commented):
Alternatively you can use goreleaser like the one i have [set up for `index-provider` CLI](https://github.com/ipni/index-provider/blob/main/.github/workflows/releaser-bin.yml). That also uploads binaries to github releases amongst other stuff; [example](https://github.com/ipni/index-provider/releases/tag/v0.10.1).

rvagg:
is goreleaser superior to this setup? I'm not really fussed as long as it works, I'd love to have consistency across projects but it seems like everyone's having a go at something different with these

masih:
I too am not fussed. For me go releaser does more with less config; that's all. Feel free to merge this as is.

rvagg:
fair enough, it is a simpler setup and the name "goreleaser" makes it feel more official...
have changed the config, ptal

.github/workflows/release-binaries.yml
```diff
@@ -0,0 +1,26 @@
+name: GoReleaser
```
masih, line 1: For goreleaser binary build to do the right thing in this repo a config like [this](https://github.com/ipni/index-provider/blob/main/.goreleaser.yaml) is needed. Once that's added try running `goreleaser check` to verify the config, and give `goreleaser --rm-dist --snapshot` a shot to make sure binaries are built as you expect them to (written in `dist` directory by default).

I also recommend adding `dist` to `.gitingore`.

masih (changes requested):

.github/workflows/release-binaries.yml
rvagg: > does more with less config
he said!

ok, added the config and tested it all out

masih: lol

masih (approved):

PR #356, opened 2023-01-27
feat: upgrade from go-block-format to go-libipfs/blocks

hacdias:
`go-block-format` is now deprecated: https://github.com/ipfs/go-block-format/pull/37

It seems like other dependencies have to be updated in order for this to work and they will also be incorporated into go-libipfs. I will update this PR once that is the case.


Jorropo:
@hacdias you just need to `go get github.com/ipfs/go-block-format@latest`, that required to make librairies like `go-ipfs-blockstore` use stubbed `go-block-format`.

Jorropo (approved):
LGTM

willscott:
tests still seem unhappy

Jorropo:
> tests still seem unhappy

Yeah there is two go.mod files and there were a typo in one of the files. Fixed it.

willscott:
3 even, right?  (v1, v2, and cmd)

Jorropo:
> 3 even, right? (v1, v2, and cmd)

Right, `cmd` imports it's own outdated version of `go-car` tho.

hacdias:
@Jorropo in a perfect world, we would also have updated https://github.com/ipfs/go-merkledag which go-car depends on.

Jorropo:
@hacdias in a perfect world `go-merkledag` and `go-car` would have been migrated to `go-libipfs` :upside_down_face: 

Using stubs is fine, open a PR on https://github.com/ipfs/go-merkledag if you want to update it.

PR #357, opened 2023-01-27
Allow using WalkOption in WriteCar function

leszko:
Pass through the `merkledag.WalkOption` parameters to the underlying `merkledag.Walk()` function.

This is helpful for a number of use-cases, in particular it allows creating a CAR with missing nodes. E.g., currently it's not possible to create a CAR with raw links pointing to CIDs not present in the local blockstore. After this change, you can use the following:

```golang
car.WriteCar(ctx, dagService, []cid.Cid{root.Cid()}, carFile, merkledag.IgnoreErrors())
```

### DIFF
--- a/car.go
+++ b/car.go
@@ -40,11 +40,11 @@ type carWriter struct {
 
 type WalkFunc func(format.Node) ([]*format.Link, error)
 
-func WriteCar(ctx context.Context, ds format.NodeGetter, roots []cid.Cid, w io.Writer) error {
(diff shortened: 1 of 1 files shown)


willscott (approved):

willscott:
I think the general sentiment has been to encourage additional feature development to happen in the v2 branch. Since this is backwards compatible / minor, and since we have continued to maintain the v1 branch I don't see any strong objection to merging.

PR #360, opened 2023-01-30
feat: add CreateWriteOnlyV1 for streamable writing blockstore

rvagg:


rvagg:
A description might be useful I guess:

```
// CreateWriteOnlyV1 creates a new WriteOnly blockstore that writes to the given
// io.Writer in CARv1 format. A WriteOnly CARv1 is able to support most of the
// Blockstore interface while also being suitable for writing to a streaming
// output as it only requires an io.Writer.
//
// The returned WriteOnly blockstore will not support Get or DeleteBlock
// operations, but keeps track of written CIDs so is able to support Has,
// GetSize and AllKeysChan operations.
```

rvagg:
Works nicely as a blockstore for graphsync: https://github.com/filecoin-project/lassie/pull/49

v2/blockstore/writeonly.go
```diff
@@ -0,0 +1,187 @@
...
+
+// CreateWriteOnlyV1 creates a new WriteOnly blockstore that writes to the given
+// io.Writer in CARv1 format. A WriteOnly CARv1 is able to support most of the
+// Blockstore interface while also being suitable for writing to a streaming
+// output as it only requires an io.Writer.
+//
+// The returned WriteOnly blockstore will not support Get or DeleteBlock
+// operations, but keeps track of written CIDs so is able to support Has,
+// GetSize and AllKeysChan operations.
+func CreateWriteOnlyV1(out io.Writer, roots []cid.Cid, opts ...carv2.Option) (*WriteOnly, error) {
```
rvagg, line 44:
```suggestion
func NewWriteOnlyV1(out io.Writer, roots []cid.Cid, opts ...carv2.Option) (*WriteOnly, error) {
```

I think this might be more idiomatic

rvagg:
Sadly not as useful as I'd hoped since this was intended for a graphsync backend and it doesn't do `LinkVisitOnlyOnce` or similar, so it'll need a `Get` to verify blocks.
Could become useful at some stage, just not for this specific case at the moment. Will leave this branch here if someone wants to take this up and champion it, the tests are all good and it all works.

PR #364, opened 2023-02-08
deps(cmd): update go-internal dependency

rvagg:
Crossing my fingers and attempting to fix the new flaky macos CI failures for testing the cmd package. I can replicate it locally after enough test re-runs and haven't been able to replicate it with this update, but it's flaky so that doesn't prove anything.

@mvdan are you aware of anything that might be coming out of testscript that's causing flaky CI executions on macos-only? We see them pretty regularly now, e.g.:

* https://github.com/ipld/go-car/actions/runs/4042032455/jobs/6949254029
* https://github.com/ipld/go-car/actions/runs/4042175271/jobs/6949563292
* https://github.com/ipld/go-car/actions/runs/4120011819/jobs/7114311099
* https://github.com/ipld/go-car/actions/runs/4045723899/jobs/6957586152

They all have the pattern:

```
[signal: killed]
          FAIL: testdata/script/root.txt:1: unexpected command failure
```

where the the script varies, i.e. it happens across different scripts, just on macos.

### DIFF
--- a/cmd/go.mod
+++ b/cmd/go.mod
@@ -16,7 +16,7 @@ require (
 	github.com/multiformats/go-multihash v0.2.1
 	github.com/multiformats/go-varint v0.0.6
 	github.com/polydawn/refmt v0.0.0-20201211092308-30ac6d18308e
-	github.com/rogpeppe/go-internal v1.8.1
+	github.com/rogpeppe/go-internal v1.9.0
 	github.com/urfave/cli/v2 v2.10.3
 	golang.org/x/exp v0.0.0-20220916125017-b168a2c6b86b
 )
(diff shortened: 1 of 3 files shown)


willscott (approved):

mvdan:
Yeah, I'm afraid that's a weird but known bug for testscript on macos-12. See https://github.com/rogpeppe/go-internal/issues/200. Downgrading to macos-11 is the only workaround I'm aware of, right now. I wrote all the details I know on that thread, so if any of you have a hunch or can investigate further on a MacOS 12 machine, that would be appreciated :)

rvagg:
great, thanks for looking @mvdan and for the ping in that thread, I'll watch progress and if I come up with any insights I'll chime in

PR #371, opened 2023-02-14
Update dependencies, upgrade unified-ci

rvagg:
Replaces #366

upgraded dependencies in v0, v2 and cmd

Also added a dependabot conifg, the only catch with that is I don't know if I should use /cmd/car or /cmd. The go.mod is in /cmd, but you can't run `go get` from in there. If nobody else knows what's the right way to do that then we can proceed as is and fix later if it doesn't update properly.


rvagg:
The macos CI failure is from https://github.com/rogpeppe/go-internal/issues/200, which is still unsolved. We're either going to have to live with it, migrate to a new tool, or figure out how to get unified-ci to downgrade our macos version in the OS matrix.

willscott (approved):

rvagg:
hat tip to @laurentsenta for the hot tip that we can use `UCI_GO_TEST_RUNNER_MACOS` as a repo variable in here to downgrade the macos runner to `"macos-11"` so go-internal doesn't keep on failing on us while https://github.com/rogpeppe/go-internal/issues/200 remains unresolved.

The trick now will be remembering that this is set so that when GitHub retires `macos-11` we don't have to scratch our heads why we're still using it here; cause it's kind of tucked away.

Issue #242, opened 2021-09-27
blockstore: ReadOnly.AllKeysChan should be able to use the index in some cases


mvdan:
I believe that if a CAR file has a MultihashIndexSorted index, and has the "fully indexed" characteristic bit, and UseWholeCIDs isn't turned on, then we could implement AllKeysChan with just the index, without having to seek through the entire inner CARv1. Because all we need is to loop over all the multihashes, including duplicates and identity CIDs, and the index contains those in the scenario laid out above.

Purely a special case optimization though. The reason I was thinking about this was in the context of the indexer provider for CARs; we will be using the index directly, so we don't strictly need this optimization, but it would be nice for blockstore users to get the optimization too.

cc @masih @willscott in case I'm missing something here.

masih:
Good point; I do however wonder could we supply a channel of CIDs from multihashes? 

The `MultihashIndexSorted` carries no information about codec. Going back to the loos specification on blockstore API, one could argue `AllKeysChan` based on it cannot supply a channel of CIDs _as they occur in CAR_.

mvdan:
If UseWholeCIDs is off (as per default), then the blockstore operates purely on multihashes, and this means AllKeysChan always returns "raw" multicodecs.

masih:
Ah so do we explicitly mention that `AllKeysChan` will supply CIDs with `Raw` codec if that option is enabled? I thought what that option means is that only _multihash_ portion is used for comparison.

mvdan:
From the UseWholeCIDs docs:

> • AllKeysChan will return the original whole CIDs, instead of with their multicodec set to "raw" to just provide multihashes.

mvdan:
And yup, this matches what go-blockstore does; the new "just multihashes" method sets multicodecs to Raw in AllKeysChan. So it's also consistent across its methods.

masih:
Perfect thank you 🍻

PR #372, opened 2023-02-14
blockstore: fast path for AllKeysChan using the index

MichaelMure:
Note: I'm not entirely sure I didn't break some assumption.

close https://github.com/ipld/go-car/issues/242

Crude benchmark:
```
func BenchmarkAllKeysChan(b *testing.B) {
	path := filepath.Join(b.TempDir(), "bench-large-v2.car")
	generateRandomCarV2File(b, path, 10<<20) // 10 MiB
	defer os.Remove(path)

	bs, err := blockstore.OpenReadWrite(path, nil)
	if err != nil {
		b.Fatal(err)
	}

	b.ReportAllocs()
	b.ResetTimer()

	for i := 0; i < b.N; i++ {
		c, err := bs.AllKeysChan(context.Background())
		if err != nil {
			b.Fatal(err)
		}
		for range c {
		}
	}
}
```

Before:
> BenchmarkAllKeysChan-12    	     772	   2022389 ns/op	   89049 B/op	    1617 allocs/op

After
> BenchmarkAllKeysChan-12    	    8824	    195779 ns/op	   30865 B/op	     642 allocs/op

10.3X faster, allocate 2.5X less memory.

### DIFF
--- a/v2/blockstore/readwrite.go
+++ b/v2/blockstore/readwrite.go
@@ -260,7 +260,38 @@ func (b *ReadWrite) Finalize() error {
 }
 
 func (b *ReadWrite) AllKeysChan(ctx context.Context) (<-chan cid.Cid, error) {
-	return b.ronly.AllKeysChan(ctx)
+	if ctx.Err() != nil {
+		return nil, ctx.Err()
+	}
+
+	b.ronly.mu.Lock()
+	defer b.ronly.mu.Unlock()
+
+	if b.ronly.closed {
+		return nil, errClosed
+	}
+
+	out := make(chan cid.Cid)
+
+	go func() {
+		defer close(out)
+		err := b.idx.ForEachCid(func(c cid.Cid, _ uint64) error {
+			if !b.opts.BlockstoreUseWholeCIDs {
+				c = cid.NewCidV1(cid.Raw, c.Hash())
+			}
+			select {
+			case out <- c:
+			case <-ctx.Done():
+				return ctx.Err()
+			}
+			return nil
+		})
+		if err != nil {
+			maybeReportError(ctx, err)
+		}
+	}()
+
+	return out, nil
 }
 
 func (b *ReadWrite) Has(ctx context.Context, key cid.Cid) (bool, error) {
--- a/v2/internal/store/insertionindex.go
+++ b/v2/internal/store/insertionindex.go
@@ -152,17 +152,23 @@ func (ii *InsertionIndex) Unmarshal(r io.Reader) error {
 }
 
 func (ii *InsertionIndex) ForEach(f func(multihash.Multihash, uint64) error) error {
-	var errr error
+	var err error
 	ii.items.AscendGreaterOrEqual(ii.items.Min(), func(i llrb.Item) bool {
 		r := i.(recordDigest).Record
-		err := f(r.Cid.Hash(), r.Offset)
-		if err != nil {
-			errr = err
-			return false
-		}
-		return true
+		err = f(r.Cid.Hash(), r.Offset)
+		return err == nil
 	})
-	return errr
+	return err
+}
+
+func (ii *InsertionIndex) ForEachCid(f func(cid.Cid, uint64) error) error {
+	var err error
+	ii.items.AscendGreaterOrEqual(ii.items.Min(), func(i llrb.Item) bool {
+		r := i.(recordDigest).Record
+		err = f(r.Cid, r.Offset)
+		return err == nil
+	})
+	return err
 }
 
 func (ii *InsertionIndex) Codec() multicodec.Code {


v2/blockstore/readwrite.go
```diff
@@ -260,7 +261,40 @@ func (b *ReadWrite) Finalize() error {
...
+		return nil, errClosed
+	}
+
+	out := make(chan cid.Cid)
+
+	go func() {
+		defer close(out)
+		err := b.idx.ForEach(func(mh multihash.Multihash, _ uint64) error {
+			select {
+			case out <- cid.NewCidV1(cid.Raw, mh):
```
rvagg, line 286: I think this probably breaks contract. It'll be fine if you're using the keys to talk back to the blockstore, but the contract doesn't specify that this is the only use of `AllKeysChan` I think, so this probably isn't OK.

But I think there is a way you can still achieve what you want. `idx` here is always `InsertionIndex` which is an internal construct that we don't leak, so you could add a `ForEachCid` to internal/store/insertionindex.go and do this loop but return the actual CIDs. Then you don't even need the `BlockstoreUseWholeCIDs` check. (I toyed with exactly this while working my way toward the eventual v2/storage implementation where it turned out I didn't need it).

We could potentially do something roughly the same in ReadOnly when we do an index generation and return an index (maybe InsertionIndex, or something else) that has a `ForEachCids` and then do a feature/type check on the index to see if we don't have to iterate because it ends up a bit strange that a ReadWrite has a fast path but ReadOnly doesn't. I won't burden you with needing to do that here if you don't want but a TODO mentioning the possibility of making ReadOnly have a fast path might be good.

MichaelMure: > I think this probably breaks contract.

Could you express that problem in the form of a go test? I don't see what I'm breaking here. Is that it's not necessarily the same codec?

MichaelMure:
@rvagg I did what you suggested, the performance boost is even more significant:

Before:
> BenchmarkAllKeysChan-12    	     772	   2022389 ns/op	   89049 B/op	    1617 allocs/op

After
> BenchmarkAllKeysChan-12    	   10000	    105874 ns/op	     144 B/op	       2 allocs/op

19.1X faster, allocate 735X less memory.

MichaelMure:
Removing the `BlockstoreUseWholeCIDs` check lead to a CI failure though, should I just put it back?

```
   === RUN   TestBlockstorePutSameHashes
      readwrite_test.go:242: expected blockstore without UseWholeCIDs to flatten on AllKeysChan
      readwrite_test.go:242: expected blockstore without UseWholeCIDs to flatten on AllKeysChan
      readwrite_test.go:242: expected blockstore without UseWholeCIDs to flatten on AllKeysChan
      readwrite_test.go:242: expected blockstore without UseWholeCIDs to flatten on AllKeysChan
      readwrite_test.go:242: expected blockstore without UseWholeCIDs to flatten on AllKeysChan
      readwrite_test.go:242: expected blockstore without UseWholeCIDs to flatten on AllKeysChan
      readwrite_test.go:242: expected blockstore without UseWholeCIDs to flatten on AllKeysChan
      readwrite_test.go:242: expected blockstore without UseWholeCIDs to flatten on AllKeysChan
      readwrite_test.go:242: expected blockstore without UseWholeCIDs to flatten on AllKeysChan
      readwrite_test.go:242: expected blockstore without UseWholeCIDs to flatten on AllKeysChan
      ```

**Comment by MichaelMure** [comment]
> Removing the BlockstoreUseWholeCIDs check lead to a CI failure though, should I just put it back?

Answering my own question, that doesn't fix that test. Yet the previous code with the multihash was passing that test ‽

I'll need someone more versed in that code to figure out if that test makes sense, and if so what should be done about it.

File: v2/blockstore/readwrite.go (hunk 1 above)
Comment (willscott) [reply v2/blockstore/readwrite.go]: i think one issue is that this always says the keys are of raw multicodec - e.g. it drops the codec part of the CID.

this means that other ipld utilities that may expect to e.g. attempt to parse dagPB to understand dag structure or otherwise care about the codec could end up with different behavior from before.

Comment (MichaelMure) [reply v2/blockstore/readwrite.go]: @willscott thanks for the explanation. FYI though, it's not in the latest code anymore.

File: v2/blockstore/readwrite.go [hunk 2]
```diff
@@ -260,7 +260,35 @@ func (b *ReadWrite) Finalize() error {
...
+	if b.ronly.closed {
+		return nil, errClosed
+	}
+
+	out := make(chan cid.Cid)
+
+	go func() {
+		defer close(out)
+		err := b.idx.ForEachCid(func(c cid.Cid, _ uint64) error {
+			select {
```
rvagg, line 279:
```suggestion
		err := b.idx.ForEachCid(func(c cid.Cid, _ uint64) error {
			if !b.opts.BlockstoreUseWholeCIDs {
				c = cid.NewCidV1(cid.Raw, c.Hash())
			}
			select {
```

rvagg: I think this is all you should need - so we have the rule: if the user requested storage with whole CIDs then they get whole (original) CIDs back from an index listing, if they just wanted multihash like a standard blockstore then they get raw CIDs out of this operation.

Essentially the same behaviour as https://github.com/ipld/go-car/blob/a2a8d2f9f60fba6cb0c0b985d9599bcc03a3e2f4/v2/blockstore/readonly.go#L408-L410

v2/internal/store/insertionindex.go
```diff
@@ -165,6 +165,20 @@ func (ii *InsertionIndex) ForEach(f func(multihash.Multihash, uint64) error) err
...
+	ii.items.AscendGreaterOrEqual(ii.items.Min(), func(i llrb.Item) bool {
+		r := i.(recordDigest).Record
+		err := f(r.Cid, r.Offset)
+		if err != nil {
+			errr = err
+			return false
+		}
+		return true
+	})
+	return errr
```
rvagg, line 179:
```suggestion
	var err error
	ii.items.AscendGreaterOrEqual(ii.items.Min(), func(i llrb.Item) bool {
		r := i.(recordDigest).Record
		err = f(r.Cid, r.Offset)
		return err == nil
	})
	return err
```

rvagg: any reason to not just use one `err`?

rvagg (changes requested):

v2/internal/store/insertionindex.go
MichaelMure: No other reason that I copied the pattern of `ForEach`. I'll fix that code as well.

MichaelMure:
@rvagg I applied your changes. Final result:
```
Before:
> BenchmarkAllKeysChan-12		885	   1256865 ns/op	   88911 B/op	    1617 allocs/op
> BenchmarkAllKeysChanUseWholeCIDs-12	1010	   1253543 ns/op	   57861 B/op	     976 allocs/op

After
> BenchmarkAllKeysChan-12		8971	    135906 ns/op	   30864 B/op	     642 allocs/op
> BenchmarkAllKeysChanUseWholeCIDs-12	13904	     86140 ns/op	     144 B/op	       2 allocs/op

BenchmarkAllKeysChan            --- 9.25X faster, allocate 2.9X less memory
BenchmarkAllKeysChanUseWholeCID --- 14.5X faster, allocate 401X less memory.
```

rvagg (approved):

PR #376, opened 2023-02-16
ReadWrite: add an alternative FinalizeReadOnly+Close flow

### LINKED ISSUE #375: Blockstore: Finalize but still readable
@MichaelMure would like to be able to `Finalize()` a CARv2 blockstore, so safely flush it to disk along with the index, but still be able to read from it.

This is reasonable and doable, but it means we either have to introduce new APIs or break the `Finalize()` contract; we probably shouldn't do that.

`Finalize()` calling `ReadOnly#Close()` is the key problem here, but if we had a separate `ReadWrite#Close()` then we could avoid that, but we'd need a separate `Finalize()`. @MichaelMure has suggested `FinalizeReadOnly()`. An alternative API he's suggested is for that call to return a closer but that's a bit of a departure from the imperative style here so I'm 👎 on that one. `Finalize()` can then be a combination of `FinalizeReadOnly()` and `Close()` and the functionality provided by `ReadOnly` can still remain intact and the `PutMany` check of `closed` could change to a check on a new `finalized` property.

@masih @willscott any objections to allowing @MichaelMure to proceed with those changes, or do you have better suggestions, or even objections?

MichaelMure:
The goal being to be able to keep reading blocks and the index after safely finalizing the file on disk. Typically for indexing purpose.

More tests are likely required, but PR for early feedback.

fix #375

### DIFF
--- a/v2/blockstore/readwrite.go
+++ b/v2/blockstore/readwrite.go
@@ -18,6 +18,10 @@ import (
 
 var _ blockstore.Blockstore = (*ReadWrite)(nil)
 
+var (
+	errFinalized = fmt.Errorf("cannot write in a carv2 blockstore after finalize")
+)
+
 // ReadWrite implements a blockstore that stores blocks in CARv2 format.
 // Blocks put into the blockstore can be read back once they are successfully written.
 // This implementation is preferable for a write-heavy workload.
@@ -34,6 +38,8 @@ type ReadWrite struct {
 	dataWriter *internalio.OffsetWriteSeeker
 	idx        *store.InsertionIndex
 	header     carv2.Header
+
+	finalized bool // also protected by ronly.mu
 
 	opts carv2.Options
 }
@@ -109,10 +115,11 @@ func OpenReadWriteFile(f *os.File, roots []cid.Cid, opts ...carv2.Option) (*Read
 	// Instantiate block store.
 	// Set the header fileld before applying options since padding options may modify header.
 	rwbs := &ReadWrite{
-		f:      f,
-		idx:    store.NewInsertionIndex(),
-		header: carv2.NewHeader(0),
-		opts:   carv2.ApplyOptions(opts...),
+		f:         f,
+		idx:       store.NewInsertionIndex(),
+		header:    carv2.NewHeader(0),
+		opts:      carv2.ApplyOptions(opts...),
+		finalized: false,
 	}
 	rwbs.ronly.opts = rwbs.opts
 
@@ -186,6 +193,9 @@ func (b *ReadWrite) PutMany(ctx context.Context, blks []blocks.Block) error {
 	if b.ronly.closed {
 		return errClosed
 	}
+	if b.finalized {
+		return errFinalized
+	}
 
 	for _, bl := range blks {
 		c := bl.Cid()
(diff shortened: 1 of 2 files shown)


v2/blockstore/readwrite.go
```diff
@@ -227,30 +237,69 @@ func (b *ReadWrite) Discard() {
...
+	b.ronly.mu.Lock()
+	defer b.ronly.mu.Unlock()
+
+	return b.closeWithoutMutex()
+}
+
+func (b *ReadWrite) closeWithoutMutex() error {
+	if !b.finalized {
+		return fmt.Errorf("called Close without FinalizeReadOnly first")
+	}
```
willscott, line 298: should we finalize in this case?

willscott (commented):
this looks fine to me in terms of structure.

A test would probably not hurt to make sure we don't break the functionality in the future

MichaelMure:
I added a test. It should be good to go, depending on what you want with https://github.com/ipld/go-car/pull/376#discussion_r1108816166.

v2/blockstore/readwrite.go
```diff
@@ -227,30 +237,70 @@ func (b *ReadWrite) Discard() {
...
+// After this call, the blockstore can no longer be used.
+func (b *ReadWrite) Close() error {
+	b.ronly.mu.Lock()
+	defer b.ronly.mu.Unlock()
+
+	return b.closeWithoutMutex()
+}
+
+func (b *ReadWrite) closeWithoutMutex() error {
+	if !b.finalized {
```
rvagg, line 297:
```suggestion
	if !b.opts.WriteAsCarV1 && !b.finalized {
```

Since it's not strictly necessary for CARv1, I think maybe we can skip being so strict in that case.

v2/blockstore/readwrite.go
rvagg: If we make this also finalise then we could eventually deprecate `Finalize()` so that `Close()` is always what you call at the end of use of a readwrite, you just have the option of also making it readonly before you do that. So I think my vote would be to add a finalisation phase in here if it's not already done, then we could update docs to make it clear you can do either.

MichaelMure: This is a bit odd to me. When working on a file, there is two outcomes: Finalize() and Discard(). Here you want, in the odd case that the caller didn't choose one, arbitrary choose Finalize()?

With the current PR, you get three clear paths:
- Discard()
- Finalize()
- FinalizeReadOnly() + Close()

If Close() also finalize, we get:
- Discard()
- Finalize()
- Close()
- FinalizeReadOnly() + Close()

I get that you might want this (although the current PR version look more obvious to me):
- Discard()
- Close()
- FinalizeReadOnly() + Close()

In that case, I'd suggest to go there in one go, and skip the confusing in-between.

v2/blockstore/readwrite.go
```diff
@@ -227,30 +237,70 @@ func (b *ReadWrite) Discard() {
...
 func (b *ReadWrite) Finalize() error {
+	b.ronly.mu.Lock()
+	defer b.ronly.mu.Unlock()
+
+	if err := b.finalizeReadOnlyWithoutMutex(); err != nil {
+		return err
+	}
+	if err := b.closeWithoutMutex(); err != nil {
+		return err
+	}
```
rvagg, line 251:
```suggestion
	for _, err := range []error{b.finalizeReadOnlyWithoutMutex(), b.closeWithoutMutex()} {
		if err != nil {
			return err
		}
	}
```

This is a change of behaviour over the existing `Finalize()` but I think it might best to at least attempt to getting closer to closing the filehandle even if we error on the way there, since we're responsible for it, leaving the user with no ability to perform any cleanup themselves.

Sorry for the late suggestion, I was just doing [something similar](https://github.com/filecoin-project/lassie/pull/124/files#diff-f20cfdf182d4cab5bbd42b104051dd0e6ff93c4482fa9c56266db9107a1fea4a) recently and this came up.

What do you think?

MichaelMure: It still looks to me like trying to catch a falling knife, but why not, you are the best to judge what make sense.

rvagg: haha, yeah, maybe so, but an error completing the index really shouldn't result in a file descriptor left dangling if we can help it

PR #381, opened 2023-02-27
Add WithTrustedCar() reader option

hsanjuan:
Attempt to make CAR traversal a little bit faster by not forcing users to hash every single block in a CAR.

### DIFF
--- a/v2/block_reader.go
+++ b/v2/block_reader.go
@@ -134,6 +134,20 @@ func (br *BlockReader) Next() (blocks.Block, error) {
 	return blocks.NewBlockWithCid(data, c)
 }
 
+// NextInsecure works in the same way as Next(), except it does not re-hash
+// every block to verify that the CID corresponds to the block. It should not
+// be used with untrusted or potentially corrupted CAR files. On the plus side, it is much ligher on the CPU.
+func (br *BlockReader) NextInsecure() (blocks.Block, error) {
+	c, data, err := util.ReadNode(br.r, br.opts.ZeroLengthSectionAsEOF, br.opts.MaxAllowedSectionSize)
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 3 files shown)


rvagg:
I think this would probably be better as an option that can be passed to `NewBlockReader()`, then we can use it elsewhere and then we don't need the extra method. Some variation on `Trusted` would probably be appropriate and match the LinkSystem language: https://github.com/ipld/go-ipld-prime/blob/02cafab33a85414419e990e143238719979445be/linking/types.go#L38

I think that this might be the only section read path that currently checks block hash matches CID; which either seems like an oversight or an unfortunate waste of resources--whichever view you want to take! Having the ability to turn on validation for each of the interfaces might be the best way to deal with the inconsistency, we'd just have to change the contract on one side of this (i.e. default to untrusted and then make blockstore and storage start validating unless told otherwise? or default to trusted and make this method skip validating unless it's told otherwise).

@willscott what's your take?

willscott:
agree in that i suspect you won't be mixing these two within a read so you can probably do it as you create the iterator rather than a separate method on the iterator

hsanjuan:
Like this? Agree it is better as an Option.

rvagg:
👌 very nice

Would you mind adding a test for this in block_reader_test.go? If you look in `TestMaxSectionLength` you'll see the manual construction of a basic CAR; I'd suggest just doing the same, but writing bytes in the section that doesn't match the CID and then running two read variations of the test (like in `TestMaxSectionLength`) - one that reads the block as-is and one that fails to read due to CID mismatch.

hsanjuan:
@rvagg there you go...

rvagg (approved):
👌

hsanjuan:
@rvagg was this ever released? We may need to add a v0.x.x tag.

hsanjuan:
oh you're right! I was somehow convinced this was a v0 feature. Sorry for the hassle

PR #384, opened 2023-03-06
feat: extract specific path, accept stdin as streaming input

rvagg:
(WIP: needs better testing, it works with the very limited test data I have at hand)

* Adds a `--path` (`-p`) to extract just a specific unixfs path (and its children)
* Adds `-` as a possible input for `--file` (`-f`) to read from stdin - the catch here is that it reads it into memory, so a large CAR is going to be a problem; but it will at least extract as the CAR blocks come in.

So `cat foo.car | car extract -f - -p /my/specific/file.txt` is a thing .. which will be even neater with `lassie fetch <cid>/my/specific/file.txt -o -`; when I make that a thing.

I think I might have also fixed the macos-12 flaky test problem, maybe it's not a github.com/rogpeppe/go-internal problem after all but related to how macos-12 handles buffered file reads.


rvagg:
Latest commit adds `-` as an argument option that will extract a _file_ to stdout. It'll error if it has to do anything more than a single file.

An alternative would be to provide `--stdout` as an argument rather than `-` since the output path is optional already and `-` is a valid (although very odd) output path that someone make want.

willscott:
this looks good! let me know when you want a full read through the code

cmd/car/extract.go
```diff
@@ -139,99 +168,144 @@ func resolvePath(root, pth string) (string, error) {
...
+		// a directory may be represented as a map of name:<link> if unixADL is applied
+		vl, err := n.AsLink()
+		if err != nil {
+			return 0, err
+		}
+		dest, err := ls.Load(ipld.LinkContext{}, vl, basicnode.Prototype.Any)
+		if err != nil {
+			if nf, ok := err.(interface{ NotFound() bool }); ok && nf.NotFound() {
+				fmt.Fprintf(c.App.ErrWriter, "data for directory entry not found: %s (skipping...)\n", name)
+				return 0, nil
```
rvagg, line 217: Highlighting this while I think about it and so it doesn't get lost in the code move—instead of erroring where a block load fails, it skips whatever it's failed to load. This needs more testing, and perhaps it's not feasable, because I've got an example where this partially works but then fails with an iterator overrun error which I don't yet understand. I'm guessing it's to do with sharded directory incompleteness but I need to work that out.

rvagg:
Added a big 'ol test script and some fixture data; ready for review @willscott

cmd/car/car.go
```diff
@@ -71,18 +71,25 @@ func main1() int {
...
+				Usage:     "Extract the contents of a car when the car encodes UnixFS data",
+				Action:    ExtractCar,
+				ArgsUsage: "[output directory|-]",
 				Flags: []cli.Flag{
 					&cli.StringFlag{
 						Name:      "file",
 						Aliases:   []string{"f"},
-						Usage:     "The car file to extract from",
+						Usage:     "The car file to extract from, or '-' to read from stdin",
 						Required:  true,
```
willscott, line 84: should we make a not specified (e.g. no -f set on cli) also be interpreted as read from stdin?

cmd/car/extract.go
```diff
@@ -33,92 +37,117 @@ func ExtractCar(c *cli.Context) error {
...
 	for _, root := range roots {
-		if err := extractRoot(c, &ls, root, outputDir); err != nil {
+		count, err := extractRoot(c, &ls, root, outputDir, path)
+		if err != nil {
 			return err
 		}
+		extractedFiles += count
+	}
+	if extractedFiles == 0 {
+		fmt.Fprintf(c.App.ErrWriter, "no files extracted\n")
```
willscott, line 79: should this be considered an error exit code?

willscott (approved):

cmd/car/extract.go
rvagg: yep, have made this a non-zero exit case in the latest push

cmd/car/car.go
rvagg: sure, see latest push for making `-f` optional and reading from stdin in that case

I'm always forgetting `-f` anyway and you get a hostile error, now you get a more gentle note that it's trying to read.

willscott (approved):

PR #387, opened 2023-03-10
blockstore: give a direct access to the index for read operations

MichaelMure:
This can be useful for advanced read-only operation, like indexing.

### DIFF
--- a/v2/blockstore/readwrite.go
+++ b/v2/blockstore/readwrite.go
@@ -10,6 +10,7 @@ import (
 	blocks "github.com/ipfs/go-libipfs/blocks"
 
 	carv2 "github.com/ipld/go-car/v2"
+	"github.com/ipld/go-car/v2/index"
 	"github.com/ipld/go-car/v2/internal/carv1"
 	"github.com/ipld/go-car/v2/internal/carv1/util"
 	internalio "github.com/ipld/go-car/v2/internal/io"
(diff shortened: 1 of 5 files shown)


v2/blockstore/readwrite.go
```diff
@@ -178,6 +179,12 @@ func (b *ReadWrite) initWithRoots(v2 bool, roots []cid.Cid) error {
 	return carv1.WriteHeader(&carv1.CarHeader{Roots: roots, Version: 1}, b.dataWriter)
 }
 
+// Index gives direct access to the index.
```
willscott, line 182: do we need this on `ReadWrite` - that's where it's dangerous.

The worry with `ReadWrite` access is you add another record, and then when you go to finalize the blockstore the extra record creeps in.

on ReadOnly, the index you access isn't going to get written back out into the underlying car file.

willscott (commented):
This seems fine to expose to me - @rvagg - do you have any major concerns?

v2/blockstore/readwrite.go
MichaelMure: I do need that on `ReadWrite`, but I can add a wrapper to enforce the read-only access.

willscott: do you need it before you finalize?
could we have [finalizeReadOnly](https://github.com/ipld/go-car/blob/master/v2/blockstore/readwrite.go#L257) give you a `ReadOnly` variant that you can then get the index from?

MichaelMure: I only need it after finalize, but I sort of see why some might.

rvagg:
I'm fine with this. Tests please, on both readonly and readwrite to prove that it's doing what you think it is.

readwrite gives you InsertionIndex which is quite a different beasty than the typical Index. A wrapper to protect write operations might be a good idea, although I tend to lean on the caveat emptor approach—if you figure out how to hack it to do your crazy thing then that's your problem; we already have plenty of warnings and safety procedures in place wrt "untrusted" indexes from third parties and this library isn't going to stop bad indexes getting published.

MichaelMure:
I added some tests.

rvagg (approved):

rvagg:
needs a rebase @MichaelMure, I'll cut a semver-minor after this gets merged

MichaelMure:
@rvagg it's rebased now.

PR #393, opened 2023-03-14
ReadWrite: faster Has() by using the in-memory index instead of reading on disk

MichaelMure:
Before:
`BenchmarkHas-12            205046            6404 ns/op            744 B/op            16 allocs/op`

After
`BenchmarkHas-12            2354449            481.3 ns/op            160 B/op            3 allocs/op`

```
func BenchmarkHas(b *testing.B) {
	ctx := context.TODO()

	path := filepath.Join(b.TempDir(), "bench-large-v2.car")
	generateRandomCarV2File(b, path, 200<<20) // 10 MiB
	defer os.Remove(path)

	subject, err := blockstore.OpenReadWrite(path, nil)

	c, err := subject.AllKeysChan(ctx)
	require.NoError(b, err)

	var allCids []cid.Cid

	for c2 := range c {
		allCids = append(allCids, c2)
	}

	b.ReportAllocs()
	b.ResetTimer()

	var idx int
	for i := 0; i < b.N; i++ {
		_, _ = subject.Has(ctx, allCids[idx])
		// require.NoError(b, err)
		// require.True(b, has)
		idx = (idx + 1) % len(allCids)
	}
}
```

### DIFF
--- a/v2/internal/store/insertionindex.go
+++ b/v2/internal/store/insertionindex.go
@@ -212,10 +212,10 @@ func (ii *InsertionIndex) Flatten(codec multicodec.Code) (index.Index, error) {
 // but it's separate as it allows us to compare Record.Cid directly,
 // whereas GetAll just provides Record.Offset.
 
-func (ii *InsertionIndex) HasExactCID(c cid.Cid) bool {
+func (ii *InsertionIndex) HasExactCID(c cid.Cid) (bool, error) {
 	d, err := multihash.Decode(c.Hash())
 	if err != nil {
-		panic(err)
+		return false, err
 	}
 	entry := recordDigest{digest: d.Digest}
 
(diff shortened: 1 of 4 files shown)


MichaelMure:
I keep falling into the same trap. `go.mod` files act as boundaries for tests, so if your run tests from the root of that package, the whole `/v2` is not tested ...

MichaelMure:
Before
> BenchmarkHas-8   	  190216	      6368 ns/op	     744 B/op	      16 allocs/op

After
> BenchmarkHas-8   	 1419169	       845.6 ns/op	     320 B/op	       6 allocs/op

v2/internal/store/insertionindex.go
```diff
@@ -237,3 +237,28 @@ func (ii *InsertionIndex) HasExactCID(c cid.Cid) bool {
 	ii.items.AscendGreaterOrEqual(entry, iter)
 	return found
 }
+
+func (ii *InsertionIndex) HasMultihash(mh multihash.Multihash) bool {
+	d, err := multihash.Decode(mh)
+	if err != nil {
+		panic(err)
```
rvagg, line 244: I can see some people not liking this; remember we're `panic()`-shy around here. It's pretty weird for it to fail at this step but it should be easy enough to just return `(bool, err)` from this function and then you can return the call in your `Has()`?

rvagg (approved):
👌

PR #396, opened 2023-03-16
blockstore: only close the file on error in OpenReadWrite, not OpenReadWriteFile

MichaelMure:
OpenReadWriteFile gives full control of the file to the caller, so it should not decide to close the file, even on error.


rvagg (approved):
Yeah, I think this is reasonable. Technically it's breaking but I think we can justify this as a fix.

PR #397, opened 2023-03-17
fix: handle (and test) WholeCID vs not; fast Has() path for storage

rvagg:
I realised too late that #393 didn't have the whole CID vs multihash differentiation in its new Has() that it needs to have. So adding it back in here and giving the same fast-path treatment to the storage API that blockstore has from #393.


rvagg:
cc @MichaelMure

willscott (approved):
this seems like a reasonable structure
the rename of store/put.go -> store/indexcheck.go makes sense

PR #409, opened 2023-04-05
fix: reading from stdin should broadcast EOF to block loaders

rvagg:
otherwise results in a blocking wait when you do an extract-all but there are missing blocks that you're waiting for a "done" signal for.


rvagg:
Ref: https://github.com/filecoin-project/filecoin-docs/pull/1812/files#r1151300008

rvagg:
Also note that `--verbose` now only writes to stderr, I'm considering that a fix.

willscott (approved):

PR #420, opened 2023-05-04
perf: reduce NewCarReader allocations

bradfitz:
This accounted for almost 10% of the garbage on one of Bluesky's server daemons (go tool pprof --alloc_space), leading to high GC CPU.

benchstat:

                         │   before    │                after                │
                         │   sec/op    │   sec/op     vs base                │
    NewCarReader_small-8   2.005µ ± 1%   1.451µ ± 5%  -27.61% (p=0.000 n=10)

                         │    before    │                after                 │
                         │     B/op     │     B/op      vs base                │
    NewCarReader_small-8   5.268Ki ± 0%   1.135Ki ± 0%  -78.46% (p=0.000 n=10)

                         │   before   │               after               │
                         │ allocs/op  │ allocs/op   vs base               │
    NewCarReader_small-8   29.00 ± 0%   27.00 ± 0%  -6.90% (p=0.000 n=10)

### DIFF
--- a/car.go
+++ b/car.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"sync"
 
 	blocks "github.com/ipfs/go-block-format"
 	cid "github.com/ipfs/go-cid"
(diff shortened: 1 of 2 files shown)


bradfitz:
cc @rvagg @whyrusleeping @bnewbold

willscott (approved):
Bluesky is probably one of the first users with lots of small car shards where this matters.

@rvagg - your eyes are probably useful here to keep tabs of what's going on in the v1 read path.

rvagg (approved):

PR #432, opened 2023-06-03
feat(cmd/car): add '--no-wrap' option to 'create' command

hacdias:
This was something I found that could be useful when creating CARs. Let me know what you think. I created a negative flag since the default is to wrap, not changing the default.

### DIFF
--- a/cmd/car/create.go
+++ b/cmd/car/create.go
@@ -31,6 +31,10 @@ func CreateCar(c *cli.Context) error {
 		return fmt.Errorf("a file destination must be specified")
 	}
 
+	if c.Bool("no-wrap") && c.Args().Len() > 1 {
+		return fmt.Errorf("no-wrap cannot be set with multiple source locations")
+	}
+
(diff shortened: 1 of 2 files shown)


cmd/car/create.go
```diff
@@ -101,6 +105,19 @@ func writeFiles(ctx context.Context, bs *blockstore.ReadWrite, paths ...string)
...
 
+	if noWrap {
+		root, _, err := builder.BuildUnixFSRecursive(paths[0], &ls)
+		if err != nil {
+			return cid.Undef, err
+		}
+		rcl, ok := root.(cidlink.Link)
+		if !ok {
+			return cid.Undef, fmt.Errorf("could not interpret %s", root)
+		}
```
willscott, line 116: is there a reason to do this as a separate block here, versus taking the `l`on line 123, setting it as `root` if the flag is set and breaking out of the for loop?

hacdias: Not really, I can change that

willscott (approved):

PR #457, opened 2023-06-22
feat: add inverse and version to filter cmd

laurentsenta:
Adds the options `--version` and `--inverse` to `car filter` so that:

```sh
echo $REMOVED_CID | car filter --version 1 --inverse ./input.car ./output.car
```

Removes the CID and outputs a carv1.

Context:
- `--inverse` is useful to filter out a CID out of a car, related: https://github.com/ipfs/gateway-conformance/issues/75
- `--version` is required to workaround https://github.com/ipfs/kubo/issues/9361

### DIFF
--- a/cmd/car/car.go
+++ b/cmd/car/car.go
@@ -116,6 +116,16 @@ func main1() int {
 						Name:  "append",
 						Usage: "Append cids to an existing output file",
 					},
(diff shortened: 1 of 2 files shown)


willscott (approved):
looks like a reasonable extension of filter

PR #490, opened 2023-08-30
Add an `AbsoluteIndexOffsets` option for readonly blockstore

willscott:
Allow opening a blockstore with an index that has offsets from the beginning of the car file, rather than from the beginning of the carv1 contained file.

Absolute indexes of this type are generated by boost. Once the car is stored to a remote location separate from the index, it's unclear if it was a v1 or v2 car. Using an absolute index means individual CIDs can be efficiently fetched without re-parses of the header.
The carv2 blockstore is used over this situation by boost for cases where a blockstore interface is needed. in this case, we would like to be able to support a mode where the reader offset for index reads doesn't offset based on the indicated start of the data section in the carv2 header.

### DIFF
--- a/v2/options.go
+++ b/v2/options.go
@@ -57,6 +57,7 @@ type Options struct {
 
 	BlockstoreAllowDuplicatePuts bool
 	BlockstoreUseWholeCIDs       bool
+	AbsoluteIndexOffsets         bool
 	MaxTraversalLinks            uint64
 	WriteAsCarV1                 bool
 	TraversalPrototypeChooser    traversal.LinkTargetNodePrototypeChooser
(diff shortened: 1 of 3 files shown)


v2/options.go
```diff
@@ -238,3 +239,12 @@ func AllowDuplicatePuts(allow bool) Option {
 		o.BlockstoreAllowDuplicatePuts = allow
 	}
 }
+
+// WithAbsoluteIndexOffsets is a read option which interprets index offsets as
+// absolute offsets from the beginning of the CARv2 file, instead of relative
+// offsets from the beginning of the Carv1 header.
+func WithAbsoluteIndexOffsets(absolute bool) Option {
```
masih, line 246: This would violate the [specification of CARv2](https://ipld.io/specs/transport/car/carv2/#index-format). Is there any chance we can capture this complexity inside boost? I am sympathetic that this might be hard to do in the existing Boost implementation, but am worried that this would open a can of worms for all sorts of CARv2 parsing errors since the specification dictates offset in index format is always from the beginning of inner CARv1. 

If not, would it make sense to add this as a characteristic in CARv2 header and spec it out that way?

rvagg:
@willscott did you close because you found a way around this or do we need to explore this further? there's probably a few options here, but if we can get by with a quick parse of the start of the file then that would be the easiest.

willscott:
because it isn't worth blocking on a discussion at the car semantics level.

Here's my ugly work-around at the boost level:
https://github.com/filecoin-project/boost/pull/1651/files#diff-701447f1a6bf0a69bba3d85f6e2cd6f23373748ea9ddfc7a6de95ebeb4dfc67fR663-R687

PR #491, opened 2023-09-01
feat: fix BlockReader#SkipNext & add SourceOffset property

rvagg:
* SkipNext's metadata calculation wasn't working properly in the case of a CARv2 as it didn't take into account the V1 header in the original offset calculation.
* Add a SourceOffset and some docs to be absolutely clear what offsets we're returning in the metadata.

Ref: https://github.com/filecoin-project/boost/pull/1651

### DIFF
--- a/v2/block_reader.go
+++ b/v2/block_reader.go
@@ -1,6 +1,7 @@
 package car
 
 import (
+	"errors"
 	"fmt"
 	"io"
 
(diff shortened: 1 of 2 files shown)


willscott (approved):
good catch.

we got lucky in that other factors have so far prevented the indexing using this method of v2 car files so far, so we don't have corrupted indices due to this issue as I understand it.

Issue #523, opened 2024-05-28
Provide Package for cmd functions


lordshashank:
We have various important functions in our [car cli](https://github.com/ipld/go-car/tree/master/cmd/car) . 
We can provide these utilities in some go-package, as would be quite helpful to people working with car files and building products. We can create package with functions like extract, etc and use them in our cli.

rvagg:
if you'd like to propose a refactoring that extracts useful functions then please do; I agree that there's a lot of useful functionality bundled up in the cmd package that could be separate—although they are mostly pretty specific to cmd so the utility package should probably be closely related to it, maybe within `cmd/`?

willscott:
@lordshashank  you mentioned `extract` as a function you'd like factored out. Are there others you'd like to see callable in the same way?

lordshashank:
Yeah so quite helpful features would be extract, root, verify, filter, inspect (returning report items if possible), and list. I mean if you are going for implementation, there would surely be many interdependent functionalities, maybe provide the important ones.  For me use cases are these, maybe someone else have others.

PR #524, opened 2024-06-05
Re-factor cmd functions to library

willscott:
Address all but `list` as requested in #523 

@lordshashank  - list seems to be really focused on a bunch of different textual printed output - i'm not really sure what a library variant of it would be vs what's already provided in the main carv2 library interface - can you specify if there's something specific you're after there?

### DIFF
--- a/cmd/car/extract.go
+++ b/cmd/car/extract.go
@@ -2,26 +2,18 @@ package main
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"io"
 	"os"
-	"path"
-	"path/filepath"
 	"runtime"
 	"strings"
 	"sync"
 
 	"github.com/ipfs/go-cid"
-	"github.com/ipfs/go-unixfsnode"
-	"github.com/ipfs/go-unixfsnode/data"
-	"github.com/ipfs/go-unixfsnode/file"
+	"github.com/ipld/go-car/cmd/car/lib"
 	"github.com/ipld/go-car/v2"
 	carstorage "github.com/ipld/go-car/v2/storage"
-	dagpb "github.com/ipld/go-codec-dagpb"
-	"github.com/ipld/go-ipld-prime"
 	cidlink "github.com/ipld/go-ipld-prime/linking/cid"
-	basicnode "github.com/ipld/go-ipld-prime/node/basic"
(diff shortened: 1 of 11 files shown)


lordshashank:
yeah I agree list command is pretty straight forward, but with unixfs it's little complex. my aim was to just get all the CIDs in a car as output from function i.e. I provide path and get the result as root cid and rest as array. Can be skipped though as root gives the root cid which is needed mostly.
Another point was, could we provide a wrapper function on extract for file and directory which just take in input and output path/locations, other function would also be available if someone wants to customize something. This is cause mostly user would just need a function that could take in their file location and give them carm not concerned about any other things.

willscott:
I don't understand the extract wrapper you're asking for - can you provide the method signature?

lordshashank:
I implemented it like this 
```

// ExtractCar pulls files and directories out of a car
func ExtractCar(carPath string, /*outputDir*/outputPath string /*, path*/) error {
	c := context.Background()

	var store storage.ReadableStorage
	var roots []cid.Cid

	carFile, err := os.Open(carPath)
	if err != nil {
		return err
	}
	store, err = carstorage.OpenReadable(carFile)
	if err != nil {
		return err
	}
	roots = store.(carstorage.ReadableCar).Roots()

	ls := cidlink.DefaultLinkSystem()
	ls.TrustedStorage = true
	ls.SetReadStorage(store)

	// path, err := pathSegments(path)
	// if err != nil {
	// 	return err
	// }

	var extractedFiles int
	for _, root := range roots {
		count, err := extractRoot(&c, &ls, root,outputPath /*outputDir, path*/)
		if err != nil {
			return err
		}
		extractedFiles += count
	}
	if extractedFiles == 0 {
		return fmt.Errorf("no files extracted")
	} else {
		fmt.Println("Extracted", extractedFiles, "files")
	}

	return nil
}
```
My need mostly was to extract files thus have commented out the directory functions (kept comments just in case need arises), basically it just takes in input and output file and gives the car, most use cases are like this only.

willscott:
Added a helper - close enough to what you're after?

lordshashank (approved):

rvagg (approved):
nice

PR #532, opened 2024-09-10
v2.14.0 bump

rvagg:


rvagg:
updated all dependencies and transitive dependencies across the 3 packages in this branch as well

rvagg:
Updated the fuzz GHA workflow to Go 1.22, also bumped the version in the fuzz files themselves which is not strictly necessary but we may as well not be tied to a specific historical compile version if it still works as is

willscott (approved):

rvagg:
binaries weren't built, https://github.com/ipld/go-car/blob/master/.github/workflows/release-binaries.yml wasn't triggered for some reason

Issue #536, opened 2024-09-19
The release workflow doesn't work


suzuki-shunsuke:
As of v2.14.0, go-car hasn't released pre built binaries.

https://github.com/ipld/go-car/releases/tag/v2.14.0
https://github.com/ipld/go-car/releases/tag/v2.14.1

<img width="333" alt="image" src="https://github.com/user-attachments/assets/2ee8d617-b6b4-48f9-9cdd-f9a68c327eaf">

Seems like the GitHub Actions workflow doesn't work.

https://github.com/ipld/go-car/actions/workflows/release-binaries.yml

<img width="802" alt="image" src="https://github.com/user-attachments/assets/dda9acb3-5f25-428a-baac-49bd2f8dccbb">

suzuki-shunsuke:
The workflow releaser.yml creates GitHub Releases.

https://github.com/ipld/go-car/blob/92d3f4e5cc6eadb2223959365be208b73d190ac7/.github/workflows/releaser.yml#L17

I think you would expect that the workflow release-binaries.yml is triggered by publishing releases, but actually the workflow wasn't triggered because releases were created by GitHub Actions token.

https://github.com/ipdxco/unified-github-workflows/blob/61bc2a929b1a896da9559d6c4b14c574425da364/.github/workflows/releaser.yml#L152C22-L152C46

GitHub Actions token doesn't trigger new workflow runs.

https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow

> When you use the repository's GITHUB_TOKEN to perform tasks, events triggered by the GITHUB_TOKEN, with the exception of workflow_dispatch and repository_dispatch, will not create a new workflow run.

So to trigger new workflow runs, you have to use a GitHub personal access token or GitHub App token instead.
The reusable workflow `ipdxco/unified-github-workflows/.github/workflows/releaser.yml` accepts a GitHub Access token `UCI_GITHUB_TOKEN`, so you can pass a GitHub personal access token but can't pass GitHub App token.

https://github.com/ipdxco/unified-github-workflows/blob/61bc2a929b1a896da9559d6c4b14c574425da364/.github/workflows/releaser.yml#L21-L23

rvagg:
[...]

PR #540, opened 2024-09-25
fix: goreleaser v2 compat, trigger release-binaries with workflow_run

rvagg:
Ref: https://github.com/ipld/go-car/issues/536

(a small amount of 🤞 going on here, we'll see, this works for lassie)

### DIFF
--- a/.github/workflows/release-binaries.yml
+++ b/.github/workflows/release-binaries.yml
@@ -2,7 +2,10 @@ name: GoReleaser
 on:
   release:
     types: [ published ]
-
+  workflow_dispatch:
+  workflow_run:
+    workflows: [Releaser]
+    types: [completed]
 jobs:
   bin-releaser:
     name: Release Binaries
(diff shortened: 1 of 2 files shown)


.github/workflows/release-binaries.yml
```diff
@@ -2,7 +2,10 @@ name: GoReleaser
 on:
   release:
     types: [ published ]
-
+  workflow_dispatch:
+  workflow_run:
+    workflows: [Releaser]
```
masih, line 7: Relases created by GitHub actions will not create other workflows, unless they use a GitHub Token other than the built-in default. 

An alternative solution is to delete and recreate the release manually.

I [pointed this out](https://filecoinproject.slack.com/archives/C03KLC57LKB/p1693986844074549) to IPDX team a while back and I believe there has been fixes rolled out to the unified CI? It's worth checking with them.

masih (approved):
Approved so that you can try the changes here if you wish. We can revert if this doesn't fix it.

.github/workflows/release-binaries.yml
rvagg: 🤷 this seems to work in the lassie repo where we have essentially the same setup and no custom token configured

rvagg: https://github.com/filecoin-project/lassie/actions/runs/9691067118

> Triggered via workflow run 3 months ago

masih: Oh nice

PR #541, opened 2024-09-25
v2.14.2 bump

rvagg:


masih (approved):

rvagg:
https://github.com/ipld/go-car/actions/runs/11027862042/job/30626938814

So, it runs, but because it depends on "releaser" completing we get it running when it shouldn't; I recall now we had this problem with lassie too, the compromise for getting this to run also involved having it run sometimes when it shouldn't. But because `Releaser` is only run on `version.json` edits, it's not as annoying as having it run on every commit, and only the creator gets the notification I think.

rvagg:
Will merge and I think it'll run again, because `Releaser` will, but this time it'll find a good tag. I guess we could put some logic into the workflow to do this test beforehand so it doesn't execute the goreleaser job.

rvagg:
https://github.com/ipld/go-car/actions/runs/11027936675

and we have binaries

so it works, but it's not ideal since we get the extra "commit isn't tagged" run that we don't need, but it's not the worst thing

PR #565, opened 2025-03-01
Add a concatination cli utility

willscott:
Usage:
`car concat a.car b.car ... > combined.car`


masih (approved):

parkan:
beautiful, we are ok with writing out the whole car in this way

PR #640, opened 2026-01-25
fix(cmd): build with local v1 and v2 mods

davidebeatrici:


davidebeatrici:
To be fair, our main concern is ensuring everything is built from source (bar a few exceptions, such as firmware blobs).

I just checked and it looks like with Go modules are never brought in as a binary but always as a source, which is really nice.

We can indeed copy the local mods into the vendor tree, but at this point it's not a problem since `go mod vendor` brings all dependencies in.

I don't have much experience with Go, so please correct me if I'm wrong.

rvagg:
That all sounds good to me, and I think the original concern here might be resolved? Correct me if I'm wrong but I'll close this for now.

davidebeatrici:
It's indeed solved, thank you again!

PR #642, opened 2026-02-09
feat/refac(cmd): Get rid of "--unixfs-blocks", introduce "--cids" and "--sizes"

davidebeatrici:
Help section with this pull request:

```
NAME:
   car list - List the blocks in a car

USAGE:
   car list [command options]

OPTIONS:
   --verbose, -v            Show verbose information about contained blocks (default: false)
   --cids                   Include CIDs (default: true except for unixfs)
   --sizes value, -s value  Include sizes (specify "human" or "bytes") (default: human for verbose, none otherwise)
   --unixfs                 Show unixfs filesystem (full paths) (default: false)
   --help, -h               show help
```

PR #653, opened 2026-02-18
feat(cmd): Make "extract" more user-friendly, matching common archive tools

davidebeatrici:
Usage:

```
car extract --file file.car [--output output] [mydir/mysubdir] [mydir/mysubdir2/file] [mydir2] [myfile]...
```

This also makes `extract` much more consistent with `create` and matching common archive tools:

- Not specifying any paths results in a complete extraction.
- To read from stdin a single dash (`-`) is specified as the input file.

### DIFF
--- a/cmd/car/car.go
+++ b/cmd/car/car.go
@@ -100,23 +100,21 @@ func main1() int {
 				}},
 			},
 			{
-				Name:      "extract",
-				Aliases:   []string{"x"},
-				Usage:     "Extract the contents of a car when the car encodes UnixFS data",
-				Action:    ExtractCar,
-				ArgsUsage: "[output directory|-]",
+				Name:    "extract",
+				Aliases: []string{"x"},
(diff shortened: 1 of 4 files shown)


davidebeatrici:
@willscott @rvagg Gentle ping, this and #642 will allow me to implement support for reading CAR files in KDE's Ark.

willscott:
I'm fine with the changes to output / specifying of partial extraction.

I think you should retain the ability to stream from stdin or pipe from output of another command. there's no reason to remove that functionality and it does not reduce consistency.  (e.g. tar / gz / etc all support extraction from stdin)

davidebeatrici:
Thank you.

Let's use the dash (`-`) to read from stdin/pipe, just like tar?

davidebeatrici:
@willscott Done.

willscott (approved):

PR #654, opened 2026-02-18
fix: remove broken trailing data check in "inspect --full" for CARv1

rvagg:
trailing-data check neither works because of the offsetReadSeeker wapping which means we're re-reading at position 0 anyway, plus malformed trailing bytes should already be caught by section read errors

Fixes: https://github.com/ipld/go-car/issues/652

This is a partial revert of dbd9059 which came in with #384 and I'm really not sure why it was added there, you can't stream to this command AFAIK, maybe I was inspecting some streamed CARs at the time and encountering something related to this.

### DIFF
--- a/cmd/car/lib/inspect.go
+++ b/cmd/car/lib/inspect.go
@@ -3,7 +3,6 @@ package lib
 import (
 	"bytes"
 	"fmt"
-	"io"
 	"os"
 	"sort"
 	"strings"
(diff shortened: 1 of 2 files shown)


willscott (approved):

rvagg:
funny that in https://github.com/ipld/go-car/pull/384 I was also fixing the cli test runner, but that time for flaky macos, now we have perma-flaky windows

PR #656, opened 2026-02-23
Support git-raw

### LINKED ISSUE #568: car get-dag gives `no decoder registered for multicodec code 120 (0x78)`
I can't seem to extract `git-raw` DAGs.

For example, if we add a Git repo to Kubo:

```
$ git init --quiet
$ mkdir -p foo/bar
$ echo hello > foo/bar/baz
$ git add foo/bar/baz 
$ git commit -m "Hi"
[master (root-commit) 4a8cc80] Hi
 1 file changed, 1 insertion(+)
 create mode 100644 foo/bar/baz
$ echo goodbye > foo/bar/baz 
$ git add foo/bar/baz 
$ git commit -m "Bye"
[master 1c641f7] Bye
 1 file changed, 1 insertion(+), 1 deletion(-)
$ ls .git/objects/*/*
.git/objects/05/4e3e01c75de2975d1bb6e436254ecf988e650f
.git/objects/1c/641f7df8a69a99c0bb1b1a1606ff1b650c6785
.git/objects/1f/72a0a3ba198abc778252eaee11142ce80244cc
.git/objects/25/42a3b2478a94f6dd2dae1a156e618a9d8bcda8
.git/objects/4a/8cc80560cf1a1a702f7bae8ec4c37aa8de9f2b
.git/objects/53/6bdc2d8a69bddedc4472fbdb3ee346d2b230a3
.git/objects/53/9ced6ac21a843e2ff88e3dc79e6ffb6dda2cc9
.git/objects/9e/0a88cb754f425b52a72ceffa96cd758c1a53ab
.git/objects/ce/013625030ba8dba906f756967f9e9ca394464a
.git/objects/dd/7e1c6f0fefe118f0b63d9f10908c460aa317a6
$ for OBJ in .git/objects/*/*; do < "$OBJ" zlib-flate -uncompress | ipfs block put --cid-codec git-raw --mhtype sha1; done
baf4bcfafjy7adr254klv2g5w4q3cktwptchgkdy
baf4bcfa4mqpx36fgtkm4boy3dilan7y3muggpbi
baf4bcfa7okqkhoqzrk6hpass5lxbcfbm5abejta
baf4bcfbfikr3er4kst3n2lnodikw4ymktwf43ka
baf4bcfckrteakygpdinhal33v2hmjq32vdpj6ky
baf4bcfctnpoc3ctjxxpnyrds7pnt5y2g2kzdbiy
baf4bcfctttwwvqq2qq7c76eohxdz4373nxnczsi
baf4bcfe6bkemw5kpijnvfjzm575jntlvrqnfhky
baf4bcfgoae3ckaylvdn2sbxxk2lh7hu4uokemsq
baf4bcfg5pyog6d7p4empbnr5t4ijbdcgbkrrpjq
```

Then make a CAR for the HEAD commit (this will include ancestors, trees and blobs thanks to the `git-raw` codec):

```
$ ipfs dag export "$(printf 'f01781114%s' "$(git rev-parse HEAD)")" > HEAD.car
 0s  986 B / ? [---------------------------=-------------------------------------------------] 54.57 KiB/s 0s 
$ car inspect HEAD.car
Version: 1
Roots: baf4bcfa4mqpx36fgtkm4boy3dilan7y3muggpbi
Root blocks present in data: Yes
Block count: 10
Min / average / max block length (bytes): 13 / 68 / 239
Min / average / max CID length (bytes): 24 / 24 / 24
Block count per codec:
	git-raw: 10
CID count per multihash:
	sha1: 10
```

`car inspect` seems happy. Now let's say we want to make a smaller CAR with just the latest tree, ignoring all the history and commit info:

```
$ car get-dag HEAD.car "$(printf 'f01781114%s' "$(git rev-parse HEAD^{tree})")" > TREE.car

2025/03/12 00:52:58 could not choose a decoder: no decoder registered for multicodec code 120 (0x78)
```

It seems like the `git-raw` multicodec isn't supported, and I couldn't find any run-time or compile-time options to adjust the supported codecs (though I might have missed it, since I'm not familiar with Go).

Warbo:
Fixes #568

The first commit follows the approach proposed in a comment on that issue, which is quite minimal but at least makes `get-dag` work.

The second commit is more invasive, but lets us `extract` git trees and blobs. This fixes [my use-case](https://github.com/Warbo/nix-helpers/blob/d6c007e9e5658aa9e72209a5192eddba0af0e6c8/helpers/fetchGitIPFS/fetchGitIPFS.nix#L250), which I've been working-around by processing individual blocks via `zlib` + `git` + `tar` + `kubo`(!)

The third commit lets `extract` work with git commit and tag objects. I don't have a direct use-case for those, but they're probably more familiar to the vast majority of git users; and trying to `extract` them using that second commit would cause a nasty error, since it assumes they're trees. This commit avoids that ambiguity by reading the git object's type from its header; but once we're doing that, it's easy enough to  `extract` tags and commits by dereferencing them to blobs/trees!

Full disclosure: I'm not a Go programmer, and used LLMs to help with the implementation. I've sanity-checked the code a few times, and made some minor improvements, so it's hopefully not vibe-coded slop ;)


cmd/go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/ipfs/go-block-format v0.2.3
 	github.com/ipfs/go-cid v0.6.0
 	github.com/ipfs/go-ipld-format v0.6.3
+	github.com/ipfs/go-ipld-git v0.1.1
```
rvagg, line 10: This is the main reason this PR is going to be a problem. Nobody cares for that package, it's got accummulated dependabot PRs that don't get merged, some of which have CVEs associated with them (not that the package is actually impacted but it becomes a transitive problem). It gets updated by web3-bot just because it's signed up for getting GitHub actions and Go version bumps, but nothing beyond that.

https://github.com/ipfs/go-ipld-git/commits/master/

davidebeatrici: From https://github.com/ipld/ipld/issues/15#issuecomment-2018956088:

> go-cid moved to multiformats. Otherwise, the focus has been on switching to go-ipld-prime, the successor to many of these libraries.

willscott: to be fair, this is only on the cmd binary and not on the library itself so i am slightly less scared about the inclusion?

Warbo: I understand the maintenance issue.
Re: "switching to go-ipld-prime", was that not done by https://github.com/ipfs/go-ipld-git/pull/46 ?

rvagg: Your call @willscott , if you're comfortable with this, I'm not a big fan because it expands the maintenance burden for people doing this out of casual interest. I'd be keen on this if there was an active team that cared about this.

PR #658, opened 2026-03-06
fix: use `%w` when wrapping errors to preserve error chain

ChayanDass:
# Fix: preserve error chain when wrapping with %w

## Summary

Use `%w` instead of `%s` when wrapping errors so that underlying errors remain in the chain. This allows callers (e.g. **IPFS Kubo**) to use `errors.Is(err, ipld.ErrNotFound{})` and to detect boxo’s “block was not found locally” and other sentinel errors.

Ref: IPFS Kubo https://github.com/ipfs/kubo/issues/10826  Kubo and other consumers need to detect `ipld.ErrNotFound` (and related “block not found” errors) through the wrapped error. With `%s`, the chain was broken and `errors.Is` could not see the original error.

### DIFF
--- a/car.go
+++ b/car.go
@@ -53,7 +53,7 @@ func WriteCarWithWalker(ctx context.Context, ds format.NodeGetter, roots []cid.C
 	}
 
 	if err := WriteHeader(h, w); err != nil {
-		return fmt.Errorf("failed to write car header: %s", err)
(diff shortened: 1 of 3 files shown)


willscott (approved):

lidel:
@rvagg any chance we can get a v2 release  with this? :pray:

rvagg:
v2.17.0 tagged

